PR #7, opened 2021-03-24
Feature/correlation api integration

Alexanderlacuna:
This PR aims to make some few changes to the api to integrate with genenetwork2


gn3/computations/correlations.py
```diff
@@ -12,6 +12,35 @@ def compute_sum(rhs: int, lhs: int)-> int:
     return rhs + lhs
 
 
+def map_shared_keys_to_values(target_sample_keys: List, target_sample_vals: dict)->List:
```
BonfaceKilz: **Nitpick**: It'd be nice to have a space around the "->"

gn3/computations/correlations.py
```diff
@@ -12,6 +12,35 @@ def compute_sum(rhs: int, lhs: int)-> int:
...
+
+    return target_dataset_data
+
+
+def map_sample_list_to_values(sample_keys: List, sample_values: List)->dict:
+    """create a dict given two arrays  example  include keys=['X1',"X2",X3]\
+    shared values =[1,2,3] expected results {X1:1,"X2":2,"X3":3}"""
+    new_dict = dict(zip(sample_keys, sample_values))
+
+    return new_dict
```
BonfaceKilz: **nitpick**: I don't think you need this function since it only does: `dict(zip(keys, values))`. You should get rid of it IMHO...

BonfaceKilz: Also, when you git rid of this; reading becomes much easier.

Reading:
```
dict(zip(["BXA", "BXD"], [1, 2, 3]))
```

is more straightforward than:

```
map_sample_list_to_values(["BXA", "BXD"], [1, 2, 3])
```

With the above, someone would have to look at the definition of `map_sample_list_to_values` which just one line! :sweat_smile:

gn3/api/correlation.py
```diff
@@ -22,11 +40,11 @@ def compute_sample_r(corr_method="pearson"):
 
     # xtodo move code below to compute_all_sampl correlation
     this_trait_data = correlation_input.get("this_trait")
-    target_datasets = correlation_input.get("target_dataset")
+    target_dataset_data = correlation_input.get("target_dataset")
```
BonfaceKilz, line 43: Something I usually do in Python when naming things I intend to remain hidden would be to start it with an underscore. I'd call `target_dataset_data` just `_dataset_data`. That's entirely a matter of style though.

gn3/api/correlation.py
```diff
@@ -39,13 +57,17 @@ def compute_lit_corr(species=None, gene_id=None):
     are fetched from the database this is the only case where the db\
     might be needed for actual computing of the correlation results"""
 
-    database_instance = mock.Mock()
+    # database_instance = mock.Mock()
```
BonfaceKilz: *issue:* Remove the commented out code. Stale comments, given enough time, and enough people working on it, can be misleading when reading the code-base.

gn3/computations/correlations.py
```diff
@@ -12,10 +12,30 @@ def compute_sum(rhs: int, lhs: int) -> int:
     return rhs + lhs
 
 
+def map_shared_keys_to_values(target_sample_keys: List, target_sample_vals: dict)-> List:
+    """function to construct target dataset data items given commoned shared\
+    keys and trait samplelist values for example given keys  >>>>>>>>>>\
+    ["BXD1", "BXD2", "BXD5", "BXD6", "BXD8", "BXD9"] and value object as\
+    "HCMA:_AT": [4.1, 5.6, 3.2, 1.1, 4.4, 2.2],TXD_AT": [6.2, 5.7, 3.6, 1.5, 4.2, 2.3]}\
+    return  results should be a list of dicts mapping the shared keys to the trait values"""
```
BonfaceKilz: *nitpick:* In docs strings, nice to start with Capitalized case. Also  no need for "\" at the end of doc-strings.

gn3/api/correlation.py
```diff
@@ -1,18 +1,36 @@
...
 from flask import jsonify
 from flask import Blueprint
 from flask import request
 
 from gn3.computations.correlations import compute_all_sample_correlation
 from gn3.computations.correlations import compute_all_lit_correlation
 from gn3.computations.correlations import compute_all_tissue_correlation
-
+from gn3.computations.correlations import map_shared_keys_to_values
+from gn3.db_utils import database_connector
```
BonfaceKilz, line 10: Hmmm... IMHO you don't need the db_utils. All it does it give back a conn right. Remember that the db params are set in a settings file right? Creating that conn object is a one liner, which I don't think needs a module of it's own. Problem with this approach is that it can lead to bloat; Makes sense?

gn3/api/correlation.py
```diff
@@ -39,13 +57,17 @@ def compute_lit_corr(species=None, gene_id=None):
     are fetched from the database this is the only case where the db\
     might be needed for actual computing of the correlation results"""
 
-    database_instance = mock.Mock()
+    # database_instance = mock.Mock()
+    database_instance, _cursor_object = database_connector()
```
BonfaceKilz: *issue*: See comment above.

gn3/api/correlation.py
```diff
@@ -39,13 +57,17 @@ def compute_lit_corr(species=None, gene_id=None):
...
     target_traits_gene_ids = request.get_json()
+    target_trait_gene_list = list(target_traits_gene_ids.items())
 
     lit_corr_results = compute_all_lit_correlation(
-        database_instance=database_instance, trait_lists=target_traits_gene_ids,
+        database_instance=database_instance, trait_lists=target_trait_gene_list,
         species=species, gene_id=gene_id)
 
+    database_instance.close()
```
BonfaceKilz: *nitpick*: Best to name "database_instance" simply "conn". My rationale is that, "conn" in most code bases is what's normally used.

PS: Naming is hard!

gn3/api/correlation.py
```diff
@@ -54,10 +76,10 @@ def compute_tissue_corr(corr_method="pearson"):
     """api endpoint fr doing tissue correlation"""
     tissue_input_data = request.get_json()
     primary_tissue_dict = tissue_input_data["primary_tissue"]
-    target_tissues_dict_list = tissue_input_data["target_tissues"]
+    target_tissues_dict = tissue_input_data["target_tissues_dict"]
```
BonfaceKilz, line 78: *question*: What happens when "target_tissues_dict" is absent? Normally, when fetching things from a dict, I prefer using "get", since you can give it has a default value of None when a key is absent. In the former case, an error is thrown which I think closes the flask server in certain cases(for example if you assumed that you had a value and tried to run something using said value).

Cc @zsloan

gn3/db_utils.py
```diff
@@ -0,0 +1,24 @@
...
+            parsed_db.password, parsed_db.path[1:])
+
+
+def database_connector()->Tuple:
+    """function to create db connector"""
+    host, user, passwd, db_name = parse_db_url()
+    conn = mdb.connect(host, user, passwd, db_name)
+    cursor = conn.cursor()
+
+    return (conn, cursor)
```
BonfaceKilz, line 23: I don't think you need this file...

gn3/db_utils.py
```diff
@@ -0,0 +1,24 @@
...
+
```
BonfaceKilz, line 24: Configure your editor to end files with a proper Unix New line. See: https://stackoverflow.com/questions/44704968/visual-studio-code-insert-new-line-at-the-end-of-files.

Alternatively, try using a superior editor(not mentioning names) for file editing :wink: :wink:

gn3/settings.py
```diff
@@ -12,6 +12,6 @@
 TMPDIR = os.environ.get("TMPDIR", tempfile.gettempdir())
 
 # SQL confs
-SQLALCHEMY_DATABASE_URI = "mysql://kabui:1234@localhost/test"
+SQL_URI = os.environ.get("SQL_URI", "mysql://kabui:1234@localhost/db_webqtl")
```
BonfaceKilz, line 15: How about, instead of using an SQL_URI, we just set the connection params. WDYT?

guix.scm
```diff
@@ -78,7 +78,7 @@
                        ("python-pylint" python-pylint)
                        ("python-redis" ,python-redis)
                        ("python-scipy" ,python-scipy)
-                       ("python-scipy" ,python-scipy)
+                       ("python-mysqlclient" ,python-mysqlclient)
```
BonfaceKilz, line 81: +1 For remembering to add this!

tests/unit/computations/test_correlation.py
```diff
@@ -46,37 +52,39 @@ def __init__(self):
...
+            return None
 
         return self.__results[0]
 
     def fetchall(self):
         """method for fetching all items from db query"""
         if self.__results is None:
-            raise IllegalOperationError()
+            return None
         return self.__results
```
BonfaceKilz, line 82: I think there's a better way to mock the conn object. If there isn't I reckon I'll extract this elsewhere since I'll use it too...

BonfaceKilz (changes requested):
@Alexanderlacuna have a look at the comments. It'd be nice if you cleaned up your comments as per the suggestions. I reckon I'll have this merged tomorrow during the day. Nice work!

tests/unit/computations/test_correlation.py
Alexanderlacuna: Actually, there is  using unittest mocks

gn3/settings.py
Alexanderlacuna: I think having env variables for each might be the only alternative cause the order matters

gn3/api/correlation.py
Alexanderlacuna: Will add a check for None results.

PR #20, opened 2021-06-14
add biweight r script and tests

Alexanderlacuna:
#### Description
Pr contains r script for biweight mid correlation

#### Questions
biweight (bicorAndPvalue function) returns a  a list with the following components, each a matrix:(bicor,p_value,z,t,no_of_observation).
for example
$bicor
          a         b
a 1.0000000 0.6702388
b 0.6702388 1.0000000

$p
            a           b
a 0.00000e+00 2.40849e-14
b 2.40849e-14 0.00000e+00

no quite sure which value to pick


gn3/computations/biweight.R
```diff
@@ -0,0 +1,23 @@
+
```
BonfaceKilz: **issue: suggestion:** Could you put this scripts in scripts/biweight.R. Then later, we could package it inside GUIX and call it from there.

gn3/computations/biweight.R
```diff
@@ -0,0 +1,23 @@
...
+library(WGCNA)
+
+arg_values <- commandArgs(trailingOnly = TRUE)
```
BonfaceKilz: @Alexanderlacuna, @zsloan , @pjotrp and @robwwilliams how about we stick to one style guide for R code. I propose: https://web.stanford.edu/class/cs109l/unrestricted/resources/google-style.html

gn3/computations/biweight.py
```diff
@@ -0,0 +1,26 @@
+
+
+"""module contains script to call biweight mid\
+correlation in R"""
+import subprocess
+import os
+from pathlib import Path
+from typing import List
+
+FILE_PATH = os.path.join(Path(__file__).parent.absolute(), "biweight.R")
```
BonfaceKilz: **suggestion:** How about setting this as a variable. Say `BIWEIGHT_RSCRIPT`. And then we could provide that from guix later? For now, we could avail that as a hard path. WDYT?

gn3/computations/biweight.py
```diff
@@ -0,0 +1,26 @@
+
+
+"""module contains script to call biweight mid\
+correlation in R"""
```
BonfaceKilz: **nitpick:** No need for the extra 2 white spaces at the top of this file. Also, with docstrings, you don't need a "\" to break to a newline :smile_cat:

gn3/computations/biweight.py
```diff
@@ -0,0 +1,26 @@
...
+from typing import List
+
+FILE_PATH = os.path.join(Path(__file__).parent.absolute(), "biweight.R")
+
+
+def call_biweight_script(trait_vals: List,
+                         target_vals: List,
+                         path_to_script: str = FILE_PATH,
+                         command: str = "Rscript"
+                         ):
```
BonfaceKilz: **suggestion:** You should probably indicate that a tuple is returned.

gn3/computations/biweight.py
```diff
@@ -0,0 +1,26 @@
...
+    '''biweight function'''
```
BonfaceKilz: **issue: nitpick:** For docstring prefer:
```
""" This is a docstring"""
```

to:
```
''' This is a docstring'''
```

gn3/computations/biweight.py
```diff
@@ -0,0 +1,26 @@
...
+correlation in R"""
+import subprocess
+import os
+from pathlib import Path
+from typing import List
+
+FILE_PATH = os.path.join(Path(__file__).parent.absolute(), "biweight.R")
+
+
+def call_biweight_script(trait_vals: List,
```
BonfaceKilz: **nitpick**: Perhaps a better name? Say "calculate_biweight_corr"

gn3/computations/correlations.py
```diff
@@ -99,8 +100,13 @@ def do_bicor(x_val, y_val) -> Tuple[float, float]:
...
     """
-    x_val, y_val = 0, 0
-    return (x_val, y_val)
+
+    try:
+        results = call_biweight_script(x_val, y_val)
+    except Exception as error:
+        raise error
+
+    return results
```
BonfaceKilz: **suggestion:**
```suggestion
    try:
        results = call_biweight_script(x_val, y_val)
        return results
    except Exception as error:
        raise error
```

I normally prefer having variables in the same block scope.

gn3/computations/test_biweight.R
```diff
@@ -0,0 +1,25 @@
+library(testthat)
```
BonfaceKilz: You could have this test inside the scripts folder I mentioned earlier.

BonfaceKilz (changes requested):
@Alexanderlacuna have a look at some of the comments. I'll merge this once you address them. ION, much thanks for this!

PR #21, opened 2021-06-23
Fix/biweight errors

Alexanderlacuna:
#### Description
this pr fixes errors while doing biweight mid correlation for example  
  - could not convert list to double use as_numeric(unlist(arg))
also to note the test for r fails testthat package cannot be found ???


BonfaceKilz:
@Alexanderlacuna much thanks for this!

PR #23, opened 2021-06-30
port correlation matrix to gn3 

Alexanderlacuna:
#### Description
- this pr  contains the code for computing principal component analysis and factor loadings and
- this pr deletes biweight scripts written in R and replaces that with pingouin
- this  pr should also contain code for  correlation matrix ported from gn2(replace classes with functions)
- adds scree plot from gn1
-  adds   tests for the above code 

todo(add tests)
assurance checks that should be done. What are the expectations -->


zsloan:
I noticed your comment about drawing the PCA figure; drawing it isn't necessary. It's just drawn with Plotly, so only the results need to be returned.

arunisaac:
Why is pandas necessary? Can't we do PCA with sklearn alone?

Alexanderlacuna:
@arunisaac  yes we can  the PCAs are  computed using sklearn  as long as you have the data in the required  format(matrix)

arunisaac:
Yes, the PCA fit function needs a matrix. But the matrix can be a regular
numpy n-dimensional array. No need for a pandas data frame. See examples in
https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html

arunisaac:
Why round the results in `do_bicor`? If it is in the interest of displaying,
it would be better to keep the precise results, and only round in the last
minute before displaying.

In `compute_eigens`, `argmax` would be more efficient than `argsort`. See
https://numpy.org/doc/stable/reference/generated/numpy.argmax.html . Also,
since you are only interested in the largest eigen value/vector, [power
iteration](https://en.wikipedia.org/wiki/Power_iteration) methods may be
faster. But, the convergence depends on the ratio of the largest eigen value
to the second-largest eigen value, and I can't seem to find implementations of
the power method for Python. So, we can think about this later.

gn3/computations/correlation_matrix.py
```diff
@@ -16,36 +16,15 @@
...
-    """compute the pca"""
+def compute_pca(matrix):
+    """compute principal component analyis"""
 
     pca = PCA()
+
     scaled_data = preprocessing.scale(matrix)
     pca.fit(scaled_data)
-
+    # generates coordinated based on the loading_scores and scaled data
```
BonfaceKilz, line 26: **nitpick:** Generally make comments in the active voice. e.g.: "Generate co-ordinates based on the loading..."

gn3/computations/correlation_matrix.py
```diff
@@ -74,29 +52,16 @@ def get_scree_plot_data(pca_obj):
     """get scree plot data for generating plot x=x_vals,y = perc_var"""
 
     perc_var = np.round(pca_obj.explained_variance_ratio_*100, decimals=1)
-    x_vals = list(range(1, len(perc_var)+1))
+
+    x_vals = [f"PC{val}" for val in list(range(1, len(perc_var)+1))]
```
BonfaceKilz, line 68: **suggestion:**:
# #
```suggestion
    x_vals = [f"PC{val}" for val in range(1, len(perc_var)+1)]
```

No need to convert the range object to a list since it's already an iterable...

BonfaceKilz:
@Alexanderlacuna What's the state of thiis PR?

Alexanderlacuna:
@BonfaceKilz  an update on this PR. It's kinda old. Not to be merged.WIll break into subunits currently, it deals with multiple items, plus it would be a good chance to refactor things.

BonfaceKilz:
Alexander Kabui ***@***.***>
anaandika:

> @BonfaceKilz an update on this PR. It's kinda
> old. Not to be merged.

Sure.  Then how about you close this PR ...

> WIll break into subunits currently, it deals
> with multiple items, plus it would be a good
> chance to refactor things.
>

... with a message indicating that it will be
worked in separate PRs?

-- 
Bonface M. K.
Free Software Activist

Alexanderlacuna:
Closing down this PR .reworked code in  separate Pr's

PR #28, opened 2021-07-22
Fix correlation sort problem

zsloan:
#### Description
Some correlation results weren't being sorted correctly. This is apparently because corr_coefficient was sometimes returned as NaN. The code checked if it was None, which meant that the NaNs got through, and the "sorted" function basically "resets" the sort every time it encounters an NaN (resulting in the final list basically having a bunch of "separate" sorted segments).

I just added something to the if statement making sure it isn't NaN.

Do regular Pearson correlations for this trait against its own dataset - https://genenetwork.org/show_trait?trait_id=ENSMUST00000035288&dataset=UTHSC-BXD-Harv_Liv_TPM_log2_1019

The top correlation should be ~1 (previously it was 0.4-something).

### DIFF
--- a/gn3/computations/correlations.py
+++ b/gn3/computations/correlations.py
@@ -1,4 +1,5 @@
 """module contains code for correlations"""
+import math
 import multiprocessing
 
 from typing import List
@@ -90,7 +91,7 @@ def compute_sample_r_correlation(trait_name, corr_method, trait_vals,
                                        target_values=sanitized_target_vals,
                                        corr_method=corr_method)
 
-        if corr_coefficient is not None:
+        if corr_coefficient is not None and not math.isnan(corr_coefficient):
             return (trait_name, corr_coefficient, p_value, num_overlap)
     return None
 


BonfaceKilz (approved):

Alexanderlacuna:
@zsloan  Looking Great. Maybe I can add tests for this.

Alexanderlacuna:
@zsloan  Am also trying to reproduce this bug could you help provide a correlation computation that generates NaN

zsloan:
When you do a sample correlation of this trait* against its own dataset (should be the default selected), there are NaNs in the full results, though those NaNs don't make it into the filtered results (so you don't see them on the output page - you only see them if you actually look at the full results returned from GN3, so the NaNs would be in the output of compute_sample_r_correlation). I only noticed them when looking at random subsets of the full results (where I noticed that sorting occurred separately before and after them).

* https://genenetwork.org/show_trait?trait_id=ENSMUST00000035288&dataset=UTHSC-BXD-Harv_Liv_TPM_log2_1019

PR #34, opened 2021-08-11
use normal function for correlation 

Alexanderlacuna:
#### Description
python multiprocessing used to speed correlation results to memory issues causing correlation 
computation to hang up 
temporary fix use normal correlation code before issue is addressed

- also rename the functions btwn those that use multiprocessing and don't

### DIFF
--- a/gn3/computations/correlations.py
+++ b/gn3/computations/correlations.py
@@ -124,11 +124,12 @@ def filter_shared_sample_keys(this_samplelist,
     return (this_vals, target_vals)
 
 
-def compute_all_sample_correlation(this_trait,
-                                   target_dataset,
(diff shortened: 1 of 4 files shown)


gn3/computations/correlations.py
```diff
@@ -124,11 +124,12 @@ def filter_shared_sample_keys(this_samplelist,
     return (this_vals, target_vals)
 
 
-def compute_all_sample_correlation(this_trait,
-                                   target_dataset,
-                                   corr_method="pearson") -> List:
+def speed_compute_all_sample_correlation(this_trait,
```
BonfaceKilz, line 127: **nitpick:** Perhaps a better name would be `fast_sample_correlation`?

PR #37, opened 2021-08-20
Heatmap generation

fredmanglis:
#### Description

Migrates the following sections of code

- Use of `neworder` variable at https://github.com/genenetwork/genenetwork1/blob/master/web/webqtl/heatmap/Heatmap.py#L120 and in the `web.weqtl.heatmap.Heatmap.draw` function
- retrieval of strains with values as at https://github.com/genenetwork/genenetwork1/blob/master/web/webqtl/heatmap/Heatmap.py#L215-221

Tests were added for these functions as the migration was happening; they can be run the usual way.

The review could also look to verify whether the migration was done correctly.

https://github.com/genenetwork/gn-gemtext-threads/blob/main/topics/gn1-migration-to-gn2/clustering.gmi

#### What are the relevant pivotal tracker stories?

https://github.com/genenetwork/gn-gemtext-threads/blob/main/topics/gn1-migration-to-gn2/clustering.gmi

#### Screenshots (if appropriate)

N/A

#### Questions

- Do we still use QTLReaper? I seem to remember the intention to move away from it.
- If not, what do we use now?


arunisaac:
**nitpick**: In the `__order_maker` function in `compute_heatmap_order`, it
would be a little more readable if we used an if-else-if ladder instead of
having independent if conditions with return statements.

**thought**: Or, even better, it would be nice if we could use pattern
matching https://www.python.org/dev/peps/pep-0636/ I don't know how good
python's pattern matching is. I'm mostly basing this suggestion based on my
experience with scheme. So, feel free to ignore this suggestion.

**suggestion**: Using a list comprehension and `zip`, we could implement
`retrieve_strains_and_values` in only one line. And, since it's only one line,
if it is not called from many different places, you might consider just
inlining that line instead of making it a separate function.

``` python
def retrieve_strains_and_values(strainlist, trait_data):
    return [(strain, value) for strain, value in zip(strainlist, trait_data) if value]
```

fredmanglis:
@arunisaac 

> nitpick: In the __order_maker function in compute_heatmap_order, it
would be a little more readable if we used an if-else-if ladder instead of
having independent if conditions with return statements.

Running the linter against that, had the linter complaining about unnecessary 'elif' after a return statement.

You might be on to something though, so I will have a look at that later. I'd written it like that so that I can consider each pathway independently of any of the other paths, and thus reduce my mental overhead while working on it.

> thought: Or, even better, it would be nice if we could use pattern
matching https://www.python.org/dev/peps/pep-0636/ I don't know how good
python's pattern matching is. I'm mostly basing this suggestion based on my
experience with scheme. So, feel free to ignore this suggestion.

Sweet! I'll have a look and see if it is a suitable replacement.

> suggestion: Using a list comprehension and zip, we could implement
retrieve_strains_and_values in only one line. And, since it's only one line,
if it is not called from many different places, you might consider just
inlining that line instead of making it a separate function.

Nice! I will have a look at this too. Thanks.

fredmanglis:
@arunisaac 

> thought: Or, even better, it would be nice if we could use pattern
matching https://www.python.org/dev/peps/pep-0636/ I don't know how good
python's pattern matching is. I'm mostly basing this suggestion based on my
experience with scheme. So, feel free to ignore this suggestion.

From the information I can glean thus far, the "Structural Pattern Matching" feature seems to be present from Python 3.10 onward, and it seems, as far as I can tell thus far, that guix has up to Python 3.8.

As such, I think this is a feature we need to keep an eye out for, but it has not quite yet made its way to our environment just yet.

arunisaac:
> Running the linter against that, had the linter complaining about
> unnecessary 'elif' after a return statement.
>
> You might be on to something though, so I will have a look at that
> later. I'd written it like that so that I can consider each pathway
> independently of any of the other paths, and thus reduce my mental
> overhead while working on it.

Makes sense! I was thinking with my lisp hat on. Lisp has no return
statement and uses a lot of cond ladders. Please feel free to do what
makes sense in Python.

arunisaac:
> From the information I can glean thus far, the "Structural Pattern
> Matching" feature seems to be present from Python 3.10 onward, and it
> seems, as far as I can tell thus far, that guix has up to Python 3.8.

Ah, ok! I guess we just have to wait.

fredmanglis:
@arunisaac @BonfaceKilz I'm having some problems ignoring the **duplicate-code** pylint errors. If anyone of you knows how to do this successfully, please add it in or point me in the appropriate direction so that we can merge this to the **main** branch as soon as possible.

BonfaceKilz:
Muriithi Frederick Muriuki
***@***.***> anaandika:

> @arunisaac @BonfaceKilz I'm having some problems
> ignoring the **duplicate-code** pylint errors. If
> anyone of you knows how to do this successfully,
> please add it in or point me in the appropriate
> direction so that we can merge this to the
> **main** branch as soon as possible.

Sure.

Add this to the top of a file:

--8<---------------cut here---------------start------------->8---
# pylint: disable=R081
--8<---------------cut here---------------end--------------->8---

For a more global way to do this, you could edit
the .pylintrc file we have and add:

--8<---------------cut here---------------start------------->8---
disable=R081
--8<---------------cut here---------------end--------------->8---

Also, I'll be looking into the pylint/ mypy errors
and the broken build system in both gn2 and gn3. I
just added an issue to track that in gemtext.

-- 
Bonface M. K. D4F09EB110177E03C28E2FE1F5BBAE1E0392253F
Free Software Activist

fredmanglis:
@BonfaceKilz @arunisaac I have fixed the linting issues, and this is now passing all checks. Please review and merge to **main** as soon as you can.

BonfaceKilz:
Muriithi Frederick Muriuki
***@***.***> anaandika:

> @BonfaceKilz @arunisaac I have fixed the linting
> issues, and this is now passing all checks. Please
> review and merge to **main** as soon as you can.

Cool!

-- 
Bonface M. K. D4F09EB110177E03C28E2FE1F5BBAE1E0392253F
Free Software Activist

gn3/api/heatmaps.py
```diff
@@ -0,0 +1,37 @@
...
+
+heatmaps = Blueprint("heatmaps", __name__)
+
+@heatmaps.route("/clustered", methods=("POST",))
+def clustered_heatmaps():
+    """
+    Parses the incoming data and responds with the JSON-serialized plotly figure
+    representing the clustered heatmap.
+    """
+    heatmap_request = request.get_json()
```
BonfaceKilz: *nitpick:* On matters naming, normally when I get data from a request, I'd rather name it to `data` as opposed which IMHO is more clearer

gn3/api/heatmaps.py
```diff
@@ -0,0 +1,37 @@
...
+    traits_names = heatmap_request.get("traits_names", tuple())
```
BonfaceKilz: *nitpick:* We only use the variable `heatmap_request` only once to get "trait_names". With naming being a chore, I would defer it here as such:
```suggestion
    traits_names = request.get_json().get("traits_names", tuple())
```

Opt out of creating variables that are only referenced once because it removes the extra cognitive load of coming up with names.

gn3/computations/qtlreaper.py
```diff
@@ -0,0 +1,169 @@
...
+computation of QTLs.
+"""
+import os
+import subprocess
+from typing import Union
+
+from gn3.random import random_string
+from gn3.settings import TMPDIR, REAPER_COMMAND
+
+def generate_traits_file(strains, trait_values, traits_filename):
```
BonfaceKilz: *nitpick:* IMHO it's nice to add types. It makes reviews easier :wink:

gn3/computations/qtlreaper.py
```diff
@@ -0,0 +1,169 @@
...
+         for t in trait_values[-1:]])
+    with open(traits_filename, "w") as outfile:
+        outfile.writelines(data)
+
+def create_output_directory(path: str):
+    """Create the output directory at `path` if it does not exist."""
+    try:
+        os.mkdir(path)
+    except OSError:
+        pass
```
BonfaceKilz: *suggestion:* Perhaps raise this instead of a simple "pass". Let the caller figure out how to deal with that. WDYT?

gn3/db/genotypes.py
```diff
@@ -0,0 +1,184 @@
...
+
+import os
+import gzip
+from typing import Union, TextIO
+
+from gn3.settings import GENOTYPE_FILES
+
+def build_genotype_file(
+        geno_name: str, base_dir: str = GENOTYPE_FILES,
+        extension: str = "geno"):
```
BonfaceKilz, line 11: *suggestion:* You could also  type-hint the return type:
```suggestion
        extension: str = "geno") -> str:
```

gn3/heatmaps.py
```diff
@@ -0,0 +1,395 @@
...
+    """
+    Get the strains and their corresponding values from `strainlist` and
+    `traits_data_list`.
+
+    This migrates the code in
+    https://github.com/genenetwork/genenetwork1/blob/master/web/webqtl/heatmap/Heatmap.py#L215-221
+    """
+    # This feels nasty! There's a lot of mutation of values here, that might
+    # indicate something untoward in the design of this function and its
+    # dependents  ==>  Review
```
BonfaceKilz, line 242: Noted!

BonfaceKilz (approved):
My comments _du jour_ are superfluous since I don't really get how heat-maps work. IMHO things seem OK to me. Only comments are:
- We'd need to be more consistent with how we document things.
- We also need to be consistent with how we do our typings

Other than that, @fredmanglis much thanks for this! I'll allow @arunisaac some time to have a look at this; then I'll merge it :wink:

zsloan:
Is there a reason why heatmaps.py isn't in the computations directory? That seems like it would be more consistent with the way correlations/mapping/etc are dealt with.

zsloan:
https://github.com/genenetwork/genenetwork3/blob/8d9bc0f29ce9208306915b079818e6f0c31785e2/gn3/db/traits.py#L422-L427

Both here and some places in heatmaps.py, "riset" should be renamed to "group"; RISet is one of the poorly named things from GN1 and was changed to "group" in GN2 (it stands for "recombinant inbred set" which doesn't make any sense for any GN data that isn't recombinant inbred, like human data).

More broadly, there might be some issues with adapting GN1 code from webqtlTrait/webqtlDataset since that code was ported over (and later changed) in GN2 (in https://github.com/genenetwork/genenetwork2/blob/testing/wqflask/base/trait.py and https://github.com/genenetwork/genenetwork2/blob/testing/wqflask/base/dataset.py respectively). It probably won't be an issue to still give GN3 the required inputs, but it would probably be best to keep stuff like naming consistent between the two projects.

zsloan:
https://github.com/genenetwork/genenetwork3/blob/8d9bc0f29ce9208306915b079818e6f0c31785e2/gn3/heatmaps.py#L168

Another example is that "strains" should instead be "samples" in places like this (with human data again being a good example where "strains" doesn't make sense).

zsloan:
Otherwise this seems fine to me; the example output shown in test_heatmaps.py is fine for passing to GN2.

My only real significant concern is that some of the ported GN1 code might be unnecessary (though it's hard for me to tell if/where since the GN1 Heatmap code is difficult for me to understand, at least from just looking at the code). The GN1 Heatmap code has the problem that it's simultaneously drawing the heatmap within the Python code. In theory, I think that the input (aside from the trait clustering, which I think just reorders the traits), should basically just be the same as "doing mapping on multiple traits."

arunisaac:
> Other than that, @fredmanglis much thanks for this! I'll allow
> @arunisaac some time to have a look at this; then I'll merge it :wink:

I haven't looked at this, but please go ahead if you think it is
fine. If you're waiting for me, it might be a few more days! ;-)

BonfaceKilz:

> https://github.com/genenetwork/genenetwork3/blob/8d9bc0f29ce9208306915b079818e6f0c31785e2/gn3/db/traits.py#L422-L427
>
> Both here and some places in heatmaps.py, "riset"
> should be renamed to "group"; RISet is one of the
> poorly named things from GN1 and was changed to
> "group" in GN2 (it stands for "recombinant inbred
> set" which doesn't make any sense for any GN data
> that isn't recombinant inbred, like human data).
>

Nice (very dominic-specific) pointer. @Fred mind
changing that? Or should we do that in another PR?

> More broadly, there might be some issues with
> adapting GN1 code from webqtlTrait/webqtlDataset
> since that code was ported over (and later
> changed) in GN2 (in
> https://github.com/genenetwork/genenetwork2/blob/testing/wqflask/base/trait.py
> and
> https://github.com/genenetwork/genenetwork2/blob/testing/wqflask/base/dataset.py
> respectively). It probably won't be an issue to
> still give GN3 the required inputs, but it would
> probably be best to keep stuff like naming
> consistent between the two projects.

-- 
Bonface M. K. D4F09EB110177E03C28E2FE1F5BBAE1E0392253F
Free Software Activist

BonfaceKilz:

>> Other than that, @fredmanglis much thanks for this! I'll allow
>> @arunisaac some time to have a look at this; then I'll merge it :wink:
>
> I haven't looked at this, but please go ahead if you think it is
> fine. If you're waiting for me, it might be a few more days! ;-)

Sure. Thanks for the heads up!

-- 
Bonface M. K. D4F09EB110177E03C28E2FE1F5BBAE1E0392253F
Free Software Activist

gn3/computations/qtlreaper.py
fredmanglis: @BonfaceKilz I have narrowed the exceptions caught in commit 60d54d8de466c179a93b6d46ad05ec1b9ba5f4a1

The point of this function is to only ever attempt to create a particular directory if it does not exist. If it does, there really is no point of passing the `FileExistsError` up to the caller. As such, I have narrowed the exceptions that are handled in the function, and let all other OSError derived exceptions through.

gn3/computations/qtlreaper.py
fredmanglis: @BonfaceKilz Yes. I add the types where I can. If adding types ends up being an impediment to moving forward, I ignore them in favour of progress, and then work toward figuring them out.

I intend to add the types by and by, especially as I gain more insight into the domain, and requirements. For now, though, please have this move forward.

gn3/api/heatmaps.py
fredmanglis: @BonfaceKilz Fixed: commit a9fc9814760d205674904f8feb700eadae480fb1

BonfaceKilz:
Muriithi Frederick Muriuki
***@***.***> anaandika:

> @fredmanglis commented on this pull request.
>
>
>
>> +    except OSError:
> +        pass
>
> @BonfaceKilz I have narrowed the exceptions caught in commit 60d54d8de466c179a93b6d46ad05ec1b9ba5f4a1
>
> The point of this function is to only ever attempt
> to create a particular directory if it does not
> exist. If it does, there really is no point of
> passing the `FileExistsError` up to the caller. As
> such, I have narrowed the exceptions that are
> handled in the function, and let all other OSError
> derived exceptions through.

Makes sense.

-- 
Bonface M. K. D4F09EB110177E03C28E2FE1F5BBAE1E0392253F
Free Software Activist

BonfaceKilz:
Muriithi Frederick Muriuki
***@***.***> anaandika:

> @fredmanglis commented on this pull request.
>
>
>
>> @@ -0,0 +1,169 @@
> +"""
> +This module contains functions to interact with the `qtlreaper` utility for
> +computation of QTLs.
> +"""
> +import os
> +import subprocess
> +from typing import Union
> +
> +from gn3.random import random_string
> +from gn3.settings import TMPDIR, REAPER_COMMAND
> +
> +def generate_traits_file(strains, trait_values, traits_filename):
>
> @BonfaceKilz Yes. I add the types where I can. If
> adding types ends up being an impediment to moving
> forward, I ignore them in favour of progress, and
> then work toward figuring them out.
>

Wisdom!

> I intend to add the types by and by, especially as
> I gain more insight into the domain, and
> requirements. For now, though, please have this
> move forward.
Sure
-- 
Bonface M. K. D4F09EB110177E03C28E2FE1F5BBAE1E0392253F
Free Software Activist

BonfaceKilz (approved):
LGTM! Going ahead to merge this.

arunisaac:
Good job, @fredmanglis and @BonfaceKilz ! Thanks for seeing this pull request through.

Alexanderlacuna:
@fredmanglis Your code looks neat. I liked the idea of how you define the inner function and private functions /'variables. This makes your codebase neat and also it is easy to group-related things without having to use OOP. Plus the tests seem to be well written and extensive.

fredmanglis:
> @fredmanglis Your code looks neat.

Thanks.

> I liked the idea of how you define the inner function and private functions /'variables. This makes your codebase neat and also it is easy to group-related things without having to use OOP.

Yes. That's a common pattern in LISPs... the trick is to not overdo it...

Take whatever lessons you can, and apply them where relevant.

> Plus the tests seem to be well written and extensive.

Sweet! I still need to improve my testing skills though.

PR #48, opened 2021-10-29
Partial correlations

fredmanglis:
#### Description

- Fix a number of issues that were leading to the tests, linter and type-checker to fail
- Rework some functions that had both computation and db-access to separate the two
- Complete the implementations of some partially implemented functions

For all, but database-accessing functions, there should be corresponding automated tests.

https://github.com/genenetwork/gn-gemtext-threads/blob/main/topics/gn1-migration-to-gn2/partial-correlations.gmi


fredmanglis:
@BonfaceKilz You can merge this.

I have even fixed the issue merged from **main** with the missing pingouin module.

BonfaceKilz:
@fredmanglis I've deleted the branch. Let me know if I should revert it...

PR #56, opened 2021-11-18
Partial correlations

fredmanglis:
Please fill out the template below, and delete items not applicable to your pull request.

#### Description

More of the partial correlation stuff.

Here, the code being migrated from GN1 written in R is replaced with calls to pingouin functions, that are better tested, and which have been probably optimised to give better performance.

**NOTE**: An attempt was made to fix the most obvious *linting* and *typing* errors. Those that I did not quite understand, I left in place to reduce the chance of conflicts when merging to main.

Run the unit tests

- https://github.com/genenetwork/gn-gemtext-threads/blob/main/topics/gn1-migration-to-gn2/partial-correlations.gmi

#### What are the relevant pivotal tracker stories?

- https://github.com/genenetwork/gn-gemtext-threads/blob/main/topics/gn1-migration-to-gn2/partial-correlations.gmi


arunisaac:
Hi Fred,

Thanks for the pull request! Everything looks good to me overall. Just a few
questions and suggestions below.

**question:** Why did you replace `assert_allclose` with
`self.assertTrue(allclose(...))`? If I understand correctly, `assert_allclose`
provides much better error messages when the test fails. `assertTrue` would
simply report that the assertion was False without saying exactly which
element of the array was off.

**nitpick:** The pingouin dependency should be the latest version
0.5.0. Version 0.3.12 is buggy, like we discussed over email. Also, it is
preferable to not specify exact versions in `requirements.txt`. It makes
packaging painful.

**suggestion (blocking):** We must never round to an arbitrary number of
decimal places. Instead tests involving floating point comparisons should
compare approximately. See my earlier commit
fb92e75b8e9984fbbf9b26f87b53ea516a8819da.

**suggestion (blocking):** In places where you have disabled pylint warnings,
could you leave a comment explaining why?

**thought:** Not a pressing issue, but as a general comment for all
genenetwork code, I think we should, as far as possible, remove the verb at
the beginning of the function name. Instead of something like
`fetch_tissue_correlations`, we can just use `tissue_correlations`. This makes
the function look more functional, and less imperative.

**nitpick:** Next time, please remove "Please fill out the template
below..."  in the pull request template.

**nitpick:** It might be better to rebase on top of main instead of merging. A
simple `git rebase main` should do the trick. Rebasing creates a more linear
commit history, and doesn't introduce new commits.

fredmanglis:
@arunisaac 

> question: Why did you replace assert_allclose with
self.assertTrue(allclose(...))? If I understand correctly, assert_allclose
provides much better error messages when the test fails. assertTrue would
simply report that the assertion was False without saying exactly which
element of the array was off.

Oh, cool. I replaced it to keep the linter quiet. I guess that can be reverted, and instead, add linter instructions to keep it quiet.


> nitpick: The pingouin dependency should be the latest version
0.5.0. Version 0.3.12 is buggy, like we discussed over email. Also, it is
preferable to not specify exact versions in requirements.txt. It makes
packaging painful.

I have not dealt with setting up the pingouin dependency myself... I guess we can just update that ourselves. WDYT?

> suggestion (blocking): We must never round to an arbitrary number of
decimal places. Instead tests involving floating point comparisons should
compare approximately. See my earlier commit

Nice! I tried looking through unittest for a similar function, and couldn't find one, so I went with the hack. I will be updating this after  the standup.

> thought: Not a pressing issue, but as a general comment for all
genenetwork code, I think we should, as far as possible, remove the verb at
the beginning of the function name. Instead of something like
fetch_tissue_correlations, we can just use tissue_correlations. This makes
the function look more functional, and less imperative.

Yeah... maybe.

I think we can retain the verbs for functions that have some side effects, and only do this for those functions that are trully functional - kinda similar to the exclamation marks in lispy dialects for functions with side-effects. WDYT?

> nitpick: It might be better to rebase on top of main instead of merging. A
simple git rebase main should do the trick. Rebasing creates a more linear
commit history, and doesn't introduce new commits.

I've never done a rebase in my career yet. :smile:  I guess one has to start somewhere...

fredmanglis:
@arunisaac I have worked on some of the items you mentioned. Have a look and merge if it is good enough.

Thanks.

zsloan:
> **thought:** Not a pressing issue, but as a general comment for all genenetwork code, I think we should, as far as possible, remove the verb at the beginning of the function name. Instead of something like `fetch_tissue_correlations`, we can just use `tissue_correlations`. This makes the function look more functional, and less imperative.

Thanks for mentioning this; this is something I'm guilty of (most examples of GN2 functions that follow this naming pattern are probably ones I'm responsible for) but what you say definitely makes more sense in most situations.

arunisaac:
>> question: Why did you replace assert_allclose with
> self.assertTrue(allclose(...))? If I understand correctly, assert_allclose
> provides much better error messages when the test fails. assertTrue would
> simply report that the assertion was False without saying exactly which
> element of the array was off.
>
> Oh, cool. I replaced it to keep the linter quiet. I guess that can be
> reverted, and instead, add linter instructions to keep it quiet.

suggestion (non-blocking): If at all possible, let's not disable the
linter check. Maybe, there is some way to address the linter check?

> I think we can retain the verbs for functions that have some side
> effects, and only do this for those functions that are trully
> functional - kinda similar to the exclamation marks in lispy dialects
> for functions with side-effects. WDYT?

My thoughts exactly! You read my mind! :-)

> I've never done a rebase in my career yet. :smile: I guess one has to
> start somewhere...

`git rebase` is a killer feature. You should definitely try it! :-) It
lets you squash/fixup commits so that the commit history is much
cleaner.

Your changes to the PR look good. Please feel free to push.

Thanks!

arunisaac:
> Thanks for mentioning this; this is something I'm guilty of (most
> examples of GN2 functions that follow this naming pattern are probably
> ones I'm responsible for) but what you say definitely makes more sense
> in most situations.

No problem, we'll gradually phase this out!

fredmanglis:
@arunisaac 

> Your changes to the PR look good. Please feel free to push.

Do you mean I should merge to main?

arunisaac:
> Do you mean I should merge to main?

Yep!

BonfaceKilz:

>> **thought:** Not a pressing issue, but as a
> general comment for all genenetwork code, I think
> we should, as far as possible, remove the verb at
> the beginning of the function name. Instead of
> something like `fetch_tissue_correlations`, we can
> just use `tissue_correlations`. This makes the
> function look more functional, and less
> imperative.

Makes sense.  Making functions a verb makes it
easy to know what side-effect is
happening. "tissue_correlations" could mean:
"fetching" correlations; or "computing"
correlations.  A general rule of thumb I
personally follow is to have functions as verbs,
and classes if any, nouns.  @Arun WDYT?

> Thanks for mentioning this; this is something I'm
> guilty of (most examples of GN2 functions that
> follow this naming pattern are probably ones I'm
> responsible for) but what you say definitely makes
> more sense in most situations.

-- 
Bonface M. K.
Free Software Activist

BonfaceKilz:
> Makes sense.  Making functions a verb makes it easy to know what
> side-effect is happening. "tissue_correlations" could mean: "fetching"
> correlations; or "computing" correlations.

Neither "fetching" nor "computing" is a side-effect. So, it shouldn't be
a verb in the function name. The function caller should not care whether
tissue correlations are fetched or computed. It should only matter that
tissue correlations are *somehow* returned. Only functions which have
side-effects, such as writing to the database, should have verbs in
their name.

arunisaac:
> Neither "fetching" nor "computing" is a side-effect. So, it shouldn't be
> a verb in the function name. The function caller should not care whether
> tissue correlations are fetched or computed. It should only matter that
> tissue correlations are *somehow* returned. Only functions which have
> side-effects, such as writing to the database, should have verbs in
> their name.

I said the above, but somehow GitHub has misattributed it to
Bonface. Looks like a GitHub bug.

fredmanglis:
> Neither "fetching" nor "computing" is a side-effect.

Hmm, I know computing is not a side-effect, but I always thought fetching, especially from say, a database, was a side-effect, kinda like I/O - one is getting data from outside the "envelope of computation".

arunisaac:
> I always thought fetching, especially from say, a database, was a
> side-effect, kinda like I/O - one is getting data from outside the
> "envelope of computation".

Maybe, I could rephrase my statement. Fetching doesn't cause
side-effects, but is affected by side-effects caused by other functions
(such as database insertions). So, given a certain database state or a
certain filesystem state, fetching is purely functional.

PR #57, opened 2021-11-19
Implement dataset metadata API endpoint.

arunisaac:
This pull request implements a dataset metadata API endpoint serving the
metadata in pages like
http://gn1.genenetwork.org/webqtl/main.py?FormID=sharinginfo&GN_AccessionId=112

This is a first draft. I am testing it extensively now. Please do review this
pull request, but do not merge it yet. Thanks!

Test with something to the effect of `curl
http://localhost:5000/api/dataset/GN112`. However this will require a running
virtuoso instance with the relevant data. We don't yet have an easily
reproducible virtuoso instance. So, you probably cannot test this.

### DIFF
--- a/gn3/api/general.py
+++ b/gn3/api/general.py
@@ -7,7 +7,7 @@ from flask import request
 
 from gn3.fs_helpers import extract_uploaded_file
 from gn3.commands import run_cmd
-
+from gn3.db import datasets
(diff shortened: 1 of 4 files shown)


arunisaac:
Pushed to master with modifications.

PR #67, opened 2021-12-22
Partial correlations fixes

fredmanglis:
This adds a number of fixes to the partial correlations feature.

Please have a look and merge to main as soon as you can.

## Extra Notes

@BonfaceKilz I think you are the best person to have a look at the following, seeing as you might have been the one working on them:

- `tests/unit/db/test_traits.py:219:29`
  The command `git blame tests/unit/db/test_traits.py | grep -v "Frederick"`
  tells me you are the only other person to work on this besides me.
  There are linting errors due to arguments/parameters to a function that have
  changed.
  Please have a look and update them correctly, so that the linter will stop
  raising errors.

- `gn3/authentication.py`
  There was a linter error about "inconsistent returns". I added a
  `return None` statement at the end to kill the linter error, but might have
  inadvertently broken something where this function is used. Please review and
  replace with the appropriate fix, and remove the comments I left you.

@arunisaac I think this is mostly yours, so please have a look and fix

- Linter errors in `gn3/db/datasets.py`
  I do not know enough to change any of this safely, so I will let you handle
  these for now

- There are also some typing issues that will need fixing when you have the time

### DIFF
--- a/gn3/api/correlation.py
+++ b/gn3/api/correlation.py
@@ -1,9 +1,9 @@
 """Endpoints for running correlations"""
 import json
-
 from flask import jsonify
 from flask import Blueprint
 from flask import request
+from flask import make_response
 
 from gn3.computations.correlations import compute_all_sample_correlation
 from gn3.computations.correlations import compute_all_lit_correlation
@@ -109,6 +109,8 @@ def partial_correlation():
         conn, trait_fullname(args["primary_trait"]),
         tuple(trait_fullname(trait) for trait in args["control_traits"]),
         args["method"], int(args["criteria"]), args["target_db"])
-    return make_response(
-        jsonify(corr_results),
-        400)
+    response = make_response(
+        json.dumps(corr_results, cls=OutputEncoder),
+        400 if "error" in corr_results.keys() else 200)
+    response.headers["Content-Type"] = "application/json"
+    return response
--- a/gn3/api/correlation.py
+++ b/gn3/api/correlation.py
@@ -1,4 +1,6 @@
 """Endpoints for running correlations"""
+import json
+
 from flask import jsonify
 from flask import Blueprint
 from flask import request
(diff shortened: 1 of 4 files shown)


arunisaac:
Thanks for the heads up, @fredmanglis ! I have fixed the linting and typing
errors I was responsible for.

pylint has a `fixme` warning that doesn't like TODO and FIXME comments, but I
think they are good practice. So, I have disabled the `fixme` warning.

In fff7954dfc7a9e843a1bca49ce0845b4f63a6342, you are replacing `NaN` with
`null` in the serialized JSON string. This is extremely dangerous. We could
easily end up replacing a valid sequence somewhere else in the string. It is
best to find and replace NaN values before the JSON tree is serialized.

Otherwise, LGTM!

fredmanglis:
@arunisaac @BonfaceKilz I have merged this.

@BonfaceKilz Please keep in mind the issues I noted above and fix them when you are able to.

arunisaac:
Hi Fred,

Did you see my comment on replacing NaN with null in the serialized JSON
string? It is critical.

Regards,
Arun

fredmanglis:
@arunisaac I did. I have been trying to figure out how to actually convert it appropriately, without changing the code too much. The only problem is that it is difficult to do this correctly without introducing an extra dependency. See [this thread](https://stackoverflow.com/questions/28639953/python-json-encoder-convert-nans-to-null-instead).

It's an annoying problem.

fredmanglis:
@arunisaac I have fixed this in commit 39779f08fd3ccf341a0855401dad1fe36119c6b4

BonfaceKilz:
Muriithi Frederick Muriuki
***@***.***> anaandika:

> @arunisaac @BonfaceKilz I have merged this.
>
> @BonfaceKilz Please keep in mind the issues I noted above and fix them when you are able to.
>
This somehow slipped in mail.  Looking at them now.
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/genenetwork/genenetwork3/pull/67#issuecomment-1000804917
> You are receiving this because you were mentioned.
>
> Message ID: ***@***.***>

-- 
Bonface M. K.
Free Software Activist

PR #68, opened 2021-12-30
Partial correlations optimisation

fredmanglis:
This just optimises the database access functions to retrieve multiple items per query, rather than a single item as they did. This improves the performance significantly, and is a first step at improving the performance of the code.

Please have a look to see whether its mostly okay to merge.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -4,4 +4,6 @@ ignore-imports=yes
 
 [MESSAGES CONTROL]
 
-disable=fixme
+disable=
+	fixme,
+	duplicate-code
--- a/gn3/api/correlation.py
+++ b/gn3/api/correlation.py
@@ -98,10 +98,10 @@ def partial_correlation():
         Class to encode output into JSON, for objects which the default
         json.JSONEncoder class does not have default encoding for.
         """
-        def default(self, obj):
-            if isinstance(obj, bytes):
-                return str(obj, encoding="utf-8")
-            return json.JSONEncoder.default(self, obj)
+        def default(self, o):
+            if isinstance(o, bytes):
+                return str(o, encoding="utf-8")
+            return json.JSONEncoder.default(self, o)
 
     args = request.get_json()
     conn, _cursor_object = database_connector()
@@ -110,7 +110,7 @@ def partial_correlation():
         tuple(trait_fullname(trait) for trait in args["control_traits"]),
         args["method"], int(args["criteria"]), args["target_db"])
     response = make_response(
-        json.dumps(corr_results, cls=OutputEncoder).replace(": NaN", ": null"),
+        json.dumps(corr_results, cls=OutputEncoder),
         400 if "error" in corr_results.keys() else 200)
     response.headers["Content-Type"] = "application/json"
     return response
--- a/gn3/computations/partial_correlations.py
+++ b/gn3/computations/partial_correlations.py
@@ -555,16 +555,13 @@ def trait_for_output(trait):
         "trait_name": trait["trait_name"],
         "symbol": trait.get("symbol"),
         "description": trait.get("description"),
-        "pre_publication_description": trait.get(
-            "pre_publication_description"),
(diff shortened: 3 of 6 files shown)


arunisaac:
In 6bb90169801671d373c86a0825d11753a76cf88d, could you add a few lines to the
commit message explaining how the commit optimizes the said function. It's not
clear to the reader.

In ada474cc84ebc98e219190c2c4d1669974481d22, I don't think we should create a
new file `gn3/computations/partial_correlations_optimised.py`. We should make
the change directly in `gn3/computations/partial_correlation.py`. That would
be much easier for comparison using git log. Creating a new file with an
"_optimised" suffix feels wrong when we have version control.

In 796416d877df5b737a3500e57d7e5f216a77f6e4, why did you change `obj` in `def
default` to the single-letter variable name `o`? Also in `__nan_to_none__`, I
think the `numpy.isnan` check alone is enough. You don't have to `or` it with
a `math.isnan` check. And, sorry to move the target away a little bit, but if
we are getting NaNs as output from correlation functions, that is symptomatic
of a much more serious bug in the correlation functions. We shouldn't be
sweeping them under the carpet by converting NaNs to Nones. Any idea what
inputs cause these NaNs? That might help us narrow down the problem.

fredmanglis:
@arunisaac 

> In 796416d, why did you change obj in def default to the single-letter variable name o?

To get rid of annoying linting error

> In ada474c, I don't think we should create a new file ...

Well, yeah. Thing is, I did the optimisation work in a separate file, then once most of the issues were handled, I got rid of the new file and merged my changes into the original file. I left this commit in place for comparison/reference. Essentially, since I rewrote history with `git rebase`, this allows review of what changed.

> Any idea what inputs cause these NaNs?

Traits with fewer than four (4) values for use in computing the partial correlations

arunisaac:
> To get rid of annoying linting error

What was this linter error? It's curious that pylint prefers to have
single-letter names..!

> Essentially, since I rewrote history with `git rebase`, this allows review
> of what changed.

Not exactly sure if it's relevant here, but in magit rebase, you can mark
commits as `fixup`. You can push these fixup commits to the pull request
branch without rewriting history. Later, when finalizing the pull request, you
can force push and clean up history.

fredmanglis:
@arunisaac 

> What was this linter error? It's curious that pylint prefers to have single-letter names..!

Nope. The linter prefers more than one character for the the names, however, the implementation of json.JSONEncoder uses a single letter for the name, which lead the linter to complain when I extended the encoder with `obj` rather than `o`

arunisaac:
Please don't let me block this pull request. Once you have addressed my
earlier comments, and are confident that everything is fine, please push to
main.

Some minor nitpicks about commit messages: Please expand commit message for
commit 4b1cf4a631ede99da0654f9b45062a28c24bc839. Also, as a general rule of
thumb, avoid starting commit messages with "Fix". For example, commit message
of 26361ace4ddcbf14b3301a81040d855bd95f3d6f could be something like "Markup
f-string with f prefix" instead of "Fix f-string". Notice how this actually
describes the change, instead of merely hinting that there has been some fix.

Cheers! :-)

BonfaceKilz:

> Please don't let me block this pull request. Once you have addressed my
> earlier comments, and are confident that everything is fine, please push to
> main.
>
> Some minor nitpicks about commit messages: Please expand commit message for
> commit 4b1cf4a631ede99da0654f9b45062a28c24bc839. Also, as a general rule of
> thumb, avoid starting commit messages with "Fix". For example, commit message
> of 26361ace4ddcbf14b3301a81040d855bd95f3d6f could be something like "Markup
> f-string with f prefix" instead of "Fix f-string". Notice how this actually
> describes the change, instead of merely hinting that there has been some fix.

Nice tip.  To add on this, I generally(but not
always) try to use the statement: "If this commit
is applied, this patch will <INSERT
COMMIT-MESSAGE>"

fredmanglis:
@arunisaac Thanks for the push-back and suggestions.

I have done some cleanup and merged the code.

PR #71, opened 2022-01-14
Use guix shell shortcut with guix.scm file

jgarte:
Is `-D` really needed in the changed command? I've been using `manifest.scm` by running `guix shell` in the directory of the manifest and all libraries seem to be available.

```
-D, --development      include the development inputs of the next package
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,15 +21,15 @@ There are at least three ways to start GeneNetwork3 with GNU Guix:
 Simply load up the environment (for development purposes):
 
 ```bash
-guix shell -Df guix.scm
+guix shell
 ```
 
 Also, make sure you have the [guix-bioinformatics](https://git.genenetwork.org/guix-bioinformatics/guix-bioinformatics) channel set up.
 
 ```bash
-guix shell --expose=$HOME/genotype_files/ -Df guix.scm
+guix shell --expose=$HOME/genotype_files/
 python3
-  import redis
+import redis
 ```
 
 #### Run a Guix container


arunisaac:
`guix shell` without the `-D` works *for now* because the genenetwork3 package
(defined in the guix-bioinformatics channel) does the funny thing of
propagating all its inputs. This is bad practice. In the future, when the
genenetwork3 package is fixed to propagate fewer or none of its inputs, `guix
shell` will cease to work without `-D`.

However, the `-f guix.scm` part of the command may not be required. Feel free
to push a commit removing that. But, please make sure to not mix it up with
indentation fixes like you are doing with the `import redis` line. That should
be a separate commit.

Thank you!

BonfaceKilz:
Just merged this.  Nothing seems off to me.  @arunisaac thanks for the insight.  Perhaps add that as an issue on gemtext threads?

arunisaac:
I meant that this commit should not be merged. We will have to undo this
commit once we fix our genenetwork3 package definition. Better to not
have merged this commit at all. May we revert it?

BonfaceKilz:

> I meant that this commit should not be merged. We will have to undo this
> commit once we fix our genenetwork3 package definition. Better to not
> have merged this commit at all. May we revert it?
>

Yeah.  Reverted here:
<https://github.com/genenetwork/genenetwork3/pull/74>

arunisaac:
Thanks, Bonface!

PR #73, opened 2022-02-01
Bug/fix csv export

zsloan:
This is intended to fix an issue where CSV exports when editing phenotypes weren't including 0 values, though in the process of looking at the issue I noticed that the CSV export wasn't working at all when using up-to-date GN3 code.

This is likely because of https://github.com/genenetwork/genenetwork3/commit/394a785da393ad659d198526f7c27a2191e90e9e

The queries are supposed to do the same thing (so deduplicating them is correct), but one had "AND PublishXRef.PhenotypeId = %s " in it while the other (retrieve_publish_trait_data) has "AND PublishFreeze.Id = %(dataset_id)s " (so after the commit it was querying for the phenotype ID in the PublishFreeze.Id column)

So I changed retrieve_publish_trait_data to instead include "AND PublishXRef.InbredSetId = %(dataset_id)s ", since the PublishFreeze table is actually unnecessary (PublishFreeze.Id = PublishXRef.InbredSetId in all but a couple situations, and in this situations you still want to use InbredSetId). There's also a separate small GN2 PR for changing the CSV URL to include the inbred_set_id instead of phenotype_id (since that's necessary now)

After this, I found what seems to be the cause of the original "0 values not showing up" issue - https://github.com/genenetwork/genenetwork3/blob/394a785da393ad659d198526f7c27a2191e90e9e/gn3/db/traits.py#L91 (the if statement evaluates as False if the value is 0, so changing it to "if record[key] is not None" fixes the issue.

There might be side effects of changing the query. I looked around and the only other place I could find that uses it is the heatmaps code, which seems to still work after the changes, but we should keep an eye out for any issues.


fredmanglis:
@zsloan 

> (the if statement evaluates as False if the value is 0, so changing it to "if record[key] is not None" fixes the issue.

Oh yes! That's a much better check. I approve.

> There might be side effects of changing the query. I looked around and the only other place I could find that uses it is the heatmaps code, which seems to still work after the changes, but we should keep an eye out for any issues.

Yes. This seems to only be used in 2 places: heatmaps in GN3, and also in GN2 where it is imported.


All the changes seem fine to me. Once @BonfaceKilz has had a look, you can probably merge

BonfaceKilz (approved):
Nothing seems off to me here.  We should start adding tests to catch regressions like this.  Easier said than done though ;)

BonfaceKilz:

> This is intended to fix an issue where CSV exports when editing phenotypes weren&#39;t including 0 values, though in the process of looking at the issue I noticed that the CSV export wasn&#39;t working at all when using up-to-date GN3 code.
>
> This is likely because of https://github.com/genenetwork/genenetwork3/commit/394a785da393ad659d198526f7c27a2191e90e9e
>

Thanks for catching this!

> The queries are supposed to do the same thing (so deduplicating them is correct), but one had &quot;AND PublishXRef.PhenotypeId = %s &quot; in it while the other (retrieve_publish_trait_data) has &quot;AND PublishFreeze.Id = %(dataset_id)s &quot; (so after the commit it was querying for the phenotype ID in the PublishFreeze.Id column)
>
> So I changed retrieve_publish_trait_data to instead include &quot;AND PublishXRef.InbredSetId = %(dataset_id)s &quot;, since the PublishFreeze table is actually unnecessary (PublishFreeze.Id = PublishXRef.InbredSetId in all but a couple situations, and in this situations you still want to use InbredSetId). There&#39;s also a separate small GN2 PR for changing the CSV URL to include the inbred_set_id instead of phenotype_id (since that&#39;s necessary now)
>
> After this, I found what seems to be the cause of the original &quot;0 values not showing up&quot; issue - https://github.com/genenetwork/genenetwork3/blob/394a785da393ad659d198526f7c27a2191e90e9e/gn3/db/traits.py#L91 (the if statement evaluates as False if the value is 0, so changing it to &quot;if record[key] is not None&quot; fixes the issue.
>
> There might be side effects of changing the query. I looked around and the only other place I could find that uses it is the heatmaps code, which seems to still work after the changes, but we should keep an eye out for any issues.

Sure!

BonfaceKilz:
FWIW @zsloan; I've just tested this; and this commit makes the downloaded CSV return wrong values; just as Suheeta confirmed on mail.

BonfaceKilz:
@zsloan Fixed this here: https://github.com/genenetwork/genenetwork3/commit/710769e84b3bc6a2bdd66effdbac0659272ed511

PR #77, opened 2022-02-09
Edit, Delete, or Add case attributes data

BonfaceKilz:
This PR:
- Extends editing Published Phenotypes to being able to also insert, delete, or edit case-attribute data.
- Adds regression tests for the above.


zsloan (approved):
This looks good to me.

One minor question - is the stuff with "longest header" just meant to get the additional CaseAttribute columns?

BonfaceKilz:

> @zsloan approved this pull request.
>
> This looks good to me.
>
> One minor question - is the stuff with "longest header" just meant to get the additional CaseAttribute columns?

This is why I avoid naming things whenever I can
;). That said, consider 2 csv files with the
following headers: "Strain Name,Value,SE,Count"
and "Strain Name,Value,SE,Count,Sex". By default,
csvdiff can't compare files with different column
headers; it will throw an error. Thus, we have to
pick the "longest" header, and on the csv file
with the shorter header, fill it with "x"s. Hope
that clarified things.


-- 
Bonface M. K.
Free Software Activist

zsloan:
Ah, that makes sense. Otherwise you'd run into an issue if a new column was
added in the updated CSV file. Thanks for explaining!

BonfaceKilz:
@zsloan I'm done with this. All the tests pass. After my class I'll fix the merge conflicts, and have this merged.

PR #79, opened 2022-02-22
Feature/refactored pca

Alexanderlacuna:
This contains refactored pca code for this PR: [https://github.com/genenetwork/genenetwork3/pull/23](https://github.com/genenetwork/genenetwork3/pull/23)


gn3/computations/principal_component_analysis.py
```diff
@@ -0,0 +1,182 @@
...
+    Returns:
+
+            coordinates(list[(x_coor,y_coord)])
+
+
+    """
+
+    perc_var = [round(ratio*100, 1) for ratio in variance_ratio]
+
+    # perc_var = np.round(variance_ratio*100, decimals=1)
```
BonfaceKilz: **nitpick**: Remove dead code.

gn3/computations/principal_component_analysis.py
```diff
@@ -0,0 +1,182 @@
...
+
+    x_coordinates = [f"PC{val}" for val in range(1, len(perc_var)+1)]
+
+    return list(zip(x_coordinates, perc_var))
```
BonfaceKilz: **issue**: Why not just return the generator and let the caller do the casting, or looping?

gn3/computations/principal_component_analysis.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+    """
+
+    for trait_id, trait_sample_data in pca_trait_dict.items():
+
+        samples_str = " ".join([str(x) for x in trait_sample_data])
+        redis_conn.set(trait_id, samples_str, ex=exp_days)
+
+    return True
```
BonfaceKilz: **issue**: Only return True for successful caches. This returns True for all caches...

tests/unit/computations/test_principal_component_analysis.py
```diff
@@ -0,0 +1,67 @@
...
+                                                   create_time="now"), expected_results)
+
+    def test_generate_scree_plot(self):
+        """test scree plot data is generated"""
+
+        variance = [0.9271, 0.06232, 0.031]
+
+        expected_results = [('PC1', 92.7), ('PC2', 6.2), ('PC3', 3.1)]
+
+        self.assertEqual(generate_scree_plot_data(variance), expected_results)
```
BonfaceKilz: **nitpick**: let's try using Pytest for all new tests. It's easier to write.

BonfaceKilz (changes requested):
work on some of the comments I've posted and "at" me for another review. Happy Hacking!

gn3/computations/principal_component_analysis.py
zsloan: @BonfaceKilz What exactly do you mean here? Would this just mean replacing `return list(zip(x_coordinates, perc_var))` with `yield zip(x_coordinates, per_var)` and calling the function like `[result for result in generate_scree_plot_data(variance_radio)]`?

zsloan (approved):
This looks good to me, aside from needing the change to the caching function that @BonfaceKilz mentioned.

gn3/computations/principal_component_analysis.py
Alexanderlacuna: made modification to this rather than returning each pair of coordinates I return the both  lists and leave  the rest for plotly
to handle

Alexanderlacuna:
@BonfaceKilz  could you review the few changes I made

BonfaceKilz (approved):
For tests, try not using classes, since with Pytest, we don't need them. See: https://github.com/genenetwork/genenetwork3/blob/main/tests/unit/test_csvcmp.py as an example.

PR #80, opened 2022-02-25
Fix issue where 0's were treated as False in correlations

zsloan:
In the function "normalize_values" in gn3/computations/correlations, there was an if statement that seems to have been evaluating 0's as False (causing them to be ignored during correlations).

I changed it to compare both a_val and b_val to None. This seems to have fixed the issue.


fredmanglis:
Nice catch! Thanks!

PR #82, opened 2022-03-08
Refactor db conn with context manager

fredmanglis:
This pull request deals with 2 main issues:

- Using context managers for database connections
- Fixing a number of issues with tests after introduction of context manager, and also pcorrs computation hand-off

Please review and LGTM


arunisaac:
LGTM, please push!

PR #105, opened 2022-11-23
Correlations: Remove `top_n` argument

fredmanglis:
The `top_n` argument is used to pick the top `n` results of the computation, but in this function, it is way too early to do that, since you need to run filters against the computation results before you pick the results.

* gn3/computations/rust_correlation.py: Remove `top_n` argument

### DIFF
--- a/gn3/computations/rust_correlation.py
+++ b/gn3/computations/rust_correlation.py
@@ -51,7 +51,7 @@ def generate_json_file(
 
 def run_correlation(
         dataset, trait_vals: str, method: str, delimiter: str,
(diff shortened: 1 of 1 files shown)


BonfaceKilz (approved):
Nothing seems off to me with this.  You could have this merged if you are happy with it.

zsloan:
The metadata fetching issue probably needs to be resolved before this is
merged; currently I think it will always result in metadata being fetched
for the entire target dataset (even if we decide to do this when using the
filtering options, the filtering options usually aren't used, so it doesn't
make sense to fetch all the metadata in those situations).

PR #117, opened 2023-05-11
Changes to sample data editing

zsloan:
This PR will make a couple changes to sample data editing:

- Values will be editable for samples that don't currently have values. Currently you can only edit sample values for samples that already have values; this is because it just pulls the sample data from the DB and wasn't accounting for the sample list
- It will be possible to submit sample data changes via an HTML table, instead of just as a CSV import

The accompanying GN2 PR is here - https://github.com/genenetwork/genenetwork2/pull/781

### DIFF
--- a/gn3/db/datasets.py
+++ b/gn3/db/datasets.py
@@ -33,6 +33,27 @@ def retrieve_sample_list(
             samplelist = headers[3:]
         return samplelist
 
+def retrieve_group_name(
+        group_id: int, connection: any):
+    """
+    Given the group id (InbredSet.Id in the database), retrieve its name
+    """
+    query = (
+        "SELECT Name "
+        "FROM InbredSet "
+        "WHERE "
+        "InbredSet.Id = %(group_id)s")
+    with connection.cursor() as cursor:
+        cursor.execute(
+            query,
+            {
+                "group_id": group_id
+            })
+        res = cursor.fetchone()
+        if res:
+            return res[0]
+        return None
+
 def retrieve_probeset_trait_dataset_name(
         threshold: int, name: str, connection: Any):
     """
--- a/gn3/db/datasets.py
+++ b/gn3/db/datasets.py
@@ -4,6 +4,35 @@ This module contains functions relating to specific trait dataset manipulation
 from typing import Any
 
 
+def retrieve_sample_list(
+        group: str, connection: Any):
+    """
+    Get the sample list for a group (a category that datasets belong to)
+
+    Currently it is fetched from the .geno files, since that's the only place
+    the "official" sample list is stored
+    """
+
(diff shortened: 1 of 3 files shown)


gn3/db/datasets.py
```diff
@@ -1,8 +1,57 @@
...
 
+def retrieve_sample_list(group: str):
+    """
+    Get the sample list for a group (a category that datasets belong to)
+
+    Currently it is fetched from the .geno files, since that's the only place
+    the "official" sample list is stored
+    """
+
+    genofile_path = os.environ.get("GENENETWORK_FILES") + "/genotype/" + group + ".geno"
```
BonfaceKilz, line 15: *nitpick:* genofile_path is a constant, and as such, upper-case it.

gn3/db/datasets.py
```diff
@@ -1,8 +1,57 @@
...
+                continue
+            break
+
+        headers = line.split("\t")
+
+        if headers[3] == "Mb":
+            samplelist = headers[4:]
+        else:
+            samplelist = headers[3:]
+        return samplelist
```
BonfaceKilz, line 33: *issue:* When opening a file, use a context-manager, in this case, `with open ...`.  This way, you don't have to worry about closing any hanging file connections should something unforeseen happens.

gn3/db/datasets.py
```diff
@@ -1,8 +1,57 @@
...
+def retrieve_group_name(
+        group_id: int, connection: any):
+    """
+    Given the group id (InbredSet.Id in the database), retrieve its name
+    """
+    query = (
+        "SELECT Name "
+        "FROM InbredSet "
+        "WHERE "
+        "InbredSet.Id = %(group_id)s")
```
BonfaceKilz, line 44: *question:* is this escaped?  I thought just using simple "%s" like so:

``` "SELECT * FROM .... WHERE ... = %s", value```

should escape things.

File: gn3/db/datasets.py [hunk 4]
```diff
@@ -1,8 +1,57 @@
...
+    with connection.cursor() as cursor:
+        cursor.execute(
+            query,
+            {
+                "group_id": group_id
+            })
+        res = cursor.fetchone()
+        if res:
+            return res[0]
+        return None
```
BonfaceKilz, line 54: *nitpick:* @arunisaac write some monadic constructs to deal with None values when doing SQL queries.  You could look at those.  If you get stuck, we could chat over on Matrix and I could show you how to use them.

BonfaceKilz (approved):
@zsloan thanks for pinging me to review this.  I've left some minor comments that you could look at.  Also, generally, you could use tests to try and poke holes at weird edge cases.

BonfaceKilz (approved):
@zsloan thanks for pinging me to review this.  I've left some minor comments that you could look at.  Also, generally, you could use tests to try and poke holes at weird edge cases.

PR #131, opened 2023-10-26
Editing case attributes

fredmanglis:
Re-implements editing case-attributes.

## Architectural & Schema Changes

- Moved the authentication code out of GN3 into [gn-auth](https://github.com/genenetwork/gn-auth). @zsloan That means running yet another service and updating configs.
- The `CaseAttribute` and `CaseAttributeXRefNew` tables will need to be updated too. The relevant script for doing the appropriate migrations is attached here, but will also be provided via a separate email.
[update-case-attribute-tables-20230818.zip](https://github.com/genenetwork/genenetwork3/files/13172227/update-case-attribute-tables-20230818.zip)

I will merge this, and work on any review feedback on main.


zsloan:
Do I need to rerun the migrations again now that case attributes are also
included)?

I'll try to get everything working with my branch tomorrow (the gnauth
service + testing the case attributes), though it might end up being a good
idea to have a meeting at some point to go over these changes (and so I can
help explain to people like Suheeta/Arthur). Would sometime early next week
work (the only time that doesn't work for me is Monday morning)? That will
give me plenty of time to try to get things working myself beforehand.

fredmanglis:
> Do I need to rerun the migrations again now that case attributes are also included)?

Yes.

To get the authorisation working for Case-Attributes, I had to do a major refactor of the auth system, which involved making changes to the db schema (auth SQLite db) so you will need to run the migrations in gn-auth.

There is also the script that I provided -- that is to be run against the MySQL database.

> I'll try to get everything working with my branch tomorrow (the gnauth
> service + testing the case attributes),

Awesome.

> though it might end up being a good idea to have a meeting at some point to go over these changes
> (and so I can help explain to people like Suheeta/Arthur).

Makes sense.

> Would sometime early next week work (the only time that doesn't work for me is Monday morning)?
> That will give me plenty of time to try to get things working myself beforehand.

Okay. It's a date then. I have added it to my calendar -- you should receive an invite in your email hopefully.

PR #136, opened 2023-10-30
Use correct upstream guix-bioinformatics channel.

BonfaceKilz:
* .guix-channel: Use git.genenetwork.org instead of gitlab.


BonfaceKilz:
Fixes [this](https://ci.genenetwork.org/jobs/genenetwork3/392)

PR #165, opened 2024-05-22
Feature/gnqa search

Alexanderlacuna:
This Pr
* replaces redis with sqlite3 for storing  gnqa search history
*  code refactoring


Alexanderlacuna:
@BonfaceKilz  When you get time can you review this work.Also look at his for related UI work:
https://github.com/genenetwork/genenetwork2/pull/847

gn3/api/llm.py
```diff
@@ -1,125 +1,113 @@
...
 from gn3.auth import db
 
+
 gnqa = Blueprint("gnqa", __name__)
 
 
-@gnqa.route("/gnqna", methods=["POST"])
-def gnqna():
-    """Main gnqa endpoint"""
+@gnqa.route("/search", methods=["POST"])
```
jnduli: minor question: why do we use POST instead of GET or PUT? I see we store the results in a table, so perhaps `PUT` may be better because its expected to be idempotent.

[ref](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST)

gn3/api/llm.py
```diff
@@ -1,125 +1,113 @@
...
-            **response,
-            "prev_queries": get_user_queries("random_user", redis_conn)
-        })
-    except LLMError as error:
-        return jsonify({"query": query,
-                        "error": f"Request failed-{str(error)}"}), 500
+    fahamu_token = current_app.config.get("FAHAMU_AUTH_TOKEN")
+    if not fahamu_token:
+        raise LLMError(
+            "Request failed:an LLM authorisation token  is required ", query)
```
jnduli, line 27: question: do we need to pass in the query to this Error? I don't think it helps understand what happened.

gn3/api/llm.py
```diff
@@ -1,125 +1,113 @@
...
+    response = {
+        "task_id": task_id,
+        "query": query,
+        "answer": answer,
+        "references": refs
+    }
+    with (db.connection(current_app.config["LLM_DB_PATH"]) as conn,
+          require_oauth.acquire("profile user") as token):
+        cursor = conn.cursor()
+        cursor.execute("""CREATE TABLE IF NOT EXISTS
```
jnduli, line 39: minor: we could have this create table queries defined as a constant at the top of the file, something like:
```
CREATE_HISTORY_TABLE = """CREATE_TABLE_IF_NOT_EXISTS ..."""
.
.
.
def search():
.
.
    cursor.execute(CREATE_HISTORY_TABLE)
```

this way, if we find a better way to ensure this modules tables are created, we can easily apply that.

gn3/api/llm.py
```diff
@@ -1,125 +1,113 @@
...
+        history(user_id TEXT NOT NULL,
+        task_id TEXT NOT NULL,
+        query  TEXT NOT NULL,
+        results  TEXT,
+        created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+        PRIMARY KEY(task_id)) WITHOUT ROWID""")
+        cursor.execute(
```
jnduli, line 46: question: do we still want to log the response if a failure happened? e.g. when we got ` "Please try to rephrase your question to receive feedback"`

gn3/api/llm.py
```diff
@@ -1,125 +1,113 @@
...
-    with (Redis.from_url(current_app.config["REDIS_URI"],
-                         decode_responses=True)) as redis_conn:
-        return jsonify({"prev_queries": get_user_queries("random_user",
-                                                         redis_conn)})
+def fetch_prev_history():
+    """Api endpoint to fetch GNQA previous search."""
+    with (require_oauth.acquire("profile user") as token,
+          db.connection(current_app.config["LLM_DB_PATH"]) as conn):
+        cursor = conn.cursor()
+        if request.method == "POST":
```
jnduli: I think this should be ` DELETE` http method

gn3/api/llm.py
```diff
@@ -1,125 +1,113 @@
...
+            and user_id=?""".format(",".join("?" * len(task_ids)))
+            cursor.execute(query, (*task_ids, str(token.user.user_id),))
+            return jsonify({})
+        if request.args.get("search_term"):
+            cursor.execute(
+                """SELECT results from history
+                Where task_id=? and user_id=?""",
+                (request.args.get("search_term"),
+                 str(token.user.user_id),))
+            return dict(cursor.fetchone())["results"]
```
jnduli: question: why do we limit this to one?

gn3/errors.py
```diff
@@ -106,6 +106,20 @@ def handle_generic(exc: Exception) -> Response:
     return resp
 
 
+def handle_llm_error(exc: Exception) -> Response:
+    """ Handle llm erros if not handled  anywhere else. """
+    resp = jsonify({
```
jnduli, line 112: can we also log this somewhere? My suggestion would be: https://github.com/jnduli/genenetwork3/blob/d4b00115b0b970b16af56f63b06cb20827695659/gn3/errors.py#L82

gn3/llms/client.py
```diff
@@ -83,44 +84,44 @@ def negative_status_msg(response):
...
-            f"HTTP error occurred  with error status:{error.response.status_code}") from error
-            except requests.exceptions.RequestException as error:
-                raise error
-        raise TimeoutError
+                    continue
+                return response
+            else:
+                raise LLMError(f"Request error with code:\
+                {response.status_code} occurred with reason:\
+                {response_msg.get(response.status_code,response.reason)}",
```
jnduli, line 122: question: can we return both? i.e. `{response_msg.get(response.status_code, "")} with their provided reason {response.reason}`

sql/update/llm_db_update.sql
```diff
@@ -0,0 +1,37 @@
+-- llm_db_update.sql ---
```
jnduli, line 1: :thinking: can we add all the create tables to this file and then find a way to create these tables any time the file changes? Note: this comment is a different approach to https://github.com/genenetwork/genenetwork3/pull/165/files#r1638117062 and I'll let you decide which may work better.

Alexanderlacuna: This makes sense. Aggregating all tables for GNQA will be added to my to-do list for today.

gn3/api/llm.py
Alexanderlacuna: there is expected to  be only one item that matches this type . What I need to do is check when an element is None before trying to access the items.

gn3/api/llm.py
Alexanderlacuna: noted.

Alexanderlacuna:
@jnduli  cc @BonfaceKilz   Thanks for the review I will be addressing the issue you raised for this PR then get back to you.

PR #167, opened 2024-06-11
Database Errors are Hidden when Running GN3

jnduli:
# Background

During set up for GN3, the home page failed loading because of a database error, but the logs didn't explain this  even with `FLASK_DEBUG=1` set. The error I got was:

```
Traceback (most recent call last):
  File "/gnu/store/1hqil50cpaq32rhl4q0p1inr25png6kx-python-flask-2.1.1/lib/python3.10/site-packages/flask/app.py", line 2095, in __call__
    return self.wsgi_app(environ, start_response)
  File "/gnu/store/1hqil50cpaq32rhl4q0p1inr25png6kx-python-flask-2.1.1/lib/python3.10/site-packages/flask/app.py", line 2080, in wsgi_app
    response = self.handle_exception(e)
  File "/gnu/store/1hqil50cpaq32rhl4q0p1inr25png6kx-python-flask-2.1.1/lib/python3.10/site-packages/flask/app.py", line 2077, in wsgi_app
    response = self.full_dispatch_request()
  File "/gnu/store/1hqil50cpaq32rhl4q0p1inr25png6kx-python-flask-2.1.1/lib/python3.10/site-packages/flask/app.py", line 1526, in full_dispatch_request
    return self.finalize_request(rv)
  File "/gnu/store/1hqil50cpaq32rhl4q0p1inr25png6kx-python-flask-2.1.1/lib/python3.10/site-packages/flask/app.py", line 1545, in finalize_request
    response = self.make_response(rv)
  File "/gnu/store/1hqil50cpaq32rhl4q0p1inr25png6kx-python-flask-2.1.1/lib/python3.10/site-packages/flask/app.py", line 1701, in make_response
    raise TypeError(
TypeError: The view function for 'menu.generate_json' did not return a valid response. The function either returned None or ended without a return statement.

```

# Solution

The `db_utils.database_connection` context manager was masking away these errors. To fix this, I've added support for logging an error here. This shouldn't change any functionality when using the module. 

# Tested

Running the same task now results in a clearer message:
```
DB error encountered
Traceback (most recent call last):
  File "/home/rookie/gn/gn3/gn3/db_utils.py", line 39, in database_connection
    yield connection
.
.
    db.query(q)
  File "/home/rookie/.guix-extra-profiles/gn3/lib/python3.10/site-packages/MySQLdb/connections.py", line 259, in query
    _mysql.connection.query(self, query)
MySQLdb._exceptions.OperationalError: (1054, "Unknown column 'Family' in 'field list'")

```

### DIFF
--- a/gn3/db_utils.py
+++ b/gn3/db_utils.py
@@ -1,6 +1,7 @@
 """module contains all db related stuff"""
 import contextlib
-from typing import Any, Iterator, Protocol, Tuple
+import logging
+from typing import Any, Iterator, Optional, Protocol, Tuple
 from urllib.parse import urlparse
 import MySQLdb as mdb
 import xapian
@@ -24,8 +25,10 @@ class Connection(Protocol):
 
 
 @contextlib.contextmanager
-def database_connection(sql_uri) -> Iterator[Connection]:
+def database_connection(sql_uri: str, logger: Optional[logging.Logger] = None) -> Iterator[Connection]:
     """Connect to MySQL database."""
+    if logger is None:
+        logger = logging.getLogger(__file__)
     host, user, passwd, db_name, port = parse_db_url(sql_uri)
     connection = mdb.connect(db=db_name,
                              user=user,
@@ -35,6 +38,7 @@ def database_connection(sql_uri) -> Iterator[Connection]:
     try:
         yield connection
     except mdb.Error as _mbde:
+        logger.error("DB error encountered", exc_info=1)
         connection.rollback()
     finally:
         connection.commit()
--- a/gn3/db_utils.py
+++ b/gn3/db_utils.py
@@ -1,10 +1,13 @@
 """module contains all db related stuff"""
 import contextlib
 import logging
-from typing import Any, Iterator, Optional, Protocol, Tuple
+from typing import Any, Iterator, Protocol, Tuple
 from urllib.parse import urlparse
 import MySQLdb as mdb
 import xapian
+
+
+LOGGER = logging.getLogger(__file__)
 
 
 def parse_db_url(sql_uri: str) -> Tuple:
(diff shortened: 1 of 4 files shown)


gn3/db_utils.py
```diff
@@ -24,8 +25,10 @@ def cursor(self, *args, **kwargs) -> Any:
 
 
 @contextlib.contextmanager
-def database_connection(sql_uri) -> Iterator[Connection]:
+def database_connection(sql_uri: str, logger: Optional[logging.Logger] = None) -> Iterator[Connection]:
     """Connect to MySQL database."""
+    if logger is None:
```
fredmanglis: Hey @jnduli 

*suggestion*: Why not move the `logger = logging.getLogger(...)` out of this function, to the module's top level, then pass it as the default logger. That way you can get rid of the `if logger is None` check and the logger is also available to any other function that might need logging in the module.

gn3/db_utils.py
```diff
@@ -35,6 +38,7 @@ def database_connection(sql_uri) -> Iterator[Connection]:
     try:
         yield connection
     except mdb.Error as _mbde:
+        logger.error("DB error encountered", exc_info=1)
```
fredmanglis, line 42: **TIL**: I missed `exc_info` as a usable argument. Nice! I have been using `traceback.format_exc` and `traceback.format_exception` when I wanted the exception's stack trace information. I'll try this one out too.

Thanks.

gn3/db_utils.py
```diff
@@ -24,8 +25,10 @@ def cursor(self, *args, **kwargs) -> Any:
 
 
 @contextlib.contextmanager
-def database_connection(sql_uri) -> Iterator[Connection]:
+def database_connection(sql_uri: str, logger: Optional[logging.Logger] = None) -> Iterator[Connection]:
```
fredmanglis: Well, if you follow my suggestion to move the creation of the logger to the module top level, you'll have to get rid of the use of `Optional` here, and line 4 above - I think that's a good (well, at the very least, acceptable) thing.

fredmanglis (changes requested):
Have a look at the reviews I have provided, make the change and you can merge it. Thanks.

fredmanglis (approved):
Hi @jnduli 

This looks good. Once you have made the changes I requested in the other commit, go right ahead and merge this.

gn3/db_utils.py
jnduli: Thanks. This makes the implementation much cleaner. I've pushed the suggestion.

gn3/db_utils.py
jnduli: You're right. This is a good thing. I've also dropped the `Optional` import.

BonfaceKilz (approved):
LGTM!

PR #169, opened 2024-06-13
Fix mypy errors in index-genenetwork script

jnduli:
## Background

Running `mypy --show-error-codes scripts/index-genenetwork` raised some errors. This PR attempts to fix the mypy errors.

I suspect the errors were due to mypy failing to check this file because it doesn't have a `.py` extenstion.

## Tested

```
╰─$ mypy --show-error-codes scripts/index-genenetwork
Success: no issues found in 1 source file

```


BonfaceKilz:
Thanks.  I'll add the index-genenetwork to the mypy tests in CI.  Thanks!

PR #175, opened 2024-07-03
Refactor how the generif md5 sum is calculated and stored in XAPIAN.

BonfaceKilz:
This PR re-works how the generif checksums are built in XAPIAN.  Instead of building it straight from virtuoso, we build the md5sum straight from the ttl file.

### DIFF
--- a/scripts/index-genenetwork
+++ b/scripts/index-genenetwork
@@ -240,7 +240,7 @@ def build_rdf_cache(sparql_uri: str, query: str, remove_common_words: bool = Fal
     return smaller_cache
 
 
-def hash_generif_graph(generif_file: str):
+def hash_generif_graph(generif_file: pathlib.Path) -> str:
     with open(generif_file, encoding="utf-8") as f_:
         data = f_.read()
         return hashlib.md5(data.encode()).hexdigest()
--- a/scripts/index-genenetwork
+++ b/scripts/index-genenetwork
@@ -24,6 +24,7 @@ import resource
 import re
 import shutil
 import sys
+import hashlib
 import tempfile
 from typing import Callable, Dict, Generator, Hashable, Iterable, List
 from SPARQLWrapper import SPARQLWrapper, JSON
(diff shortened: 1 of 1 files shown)


scripts/index-genenetwork
```diff
@@ -549,7 +525,11 @@ def is_data_modified(xapian_directory: str,
             ])
         # Return a zero exit status code when the data has changed;
         # otherwise exit with a 1 exit status code.
-        if (db.get_metadata("generif-checksum").decode() == hash_generif_graph(sparql_uri) and
+        generif = pathlib.Path("/var/lib/data/generif-metadata.ttl")
```
jnduli, line 532: hardcoding this here makes it impossible to test locally or if our system ever changes storage paths. Can we pass this in as a cmd arg?

scripts/index-genenetwork
```diff
@@ -239,35 +240,10 @@ def build_rdf_cache(sparql_uri: str, query: str, remove_common_words: bool = Fal
...
-   }}
-   } GROUP BY ?type
-"""
-    sparql.setQuery(query)
-    results = sparql.queryAndConvert()
-    if not isinstance(results, dict):
-        raise TypeError(f"Expected results to be a dict but found {type(results)}")
-    bindings = results["results"]["bindings"]
-    return bindings[0]["hash"]["value"]
+def hash_generif_graph(generif_file: str):
```
jnduli: 1. minor: add typehints for the return type
2. minor: change generif_file type to pathlib.Path
2. can we instead pass in a folder and generate the checksum for all ttl files there? This way, if we ever need to add an index for another file, we don't need to touch checksums

That would be something like:

```
ttl_hash = hashlib.new("md5")
for file in generif_file.glob("*.ttl"):
   with open(file, encoding="utf-8") as f:
       data = f.read()
        ttl_hash.update(data)
return ttl_hash.hexdigest()
```

scripts/index-genenetwork
```diff
@@ -549,7 +525,11 @@ def is_data_modified(xapian_directory: str,
             ])
         # Return a zero exit status code when the data has changed;
         # otherwise exit with a 1 exit status code.
-        if (db.get_metadata("generif-checksum").decode() == hash_generif_graph(sparql_uri) and
+        generif = pathlib.Path("/var/lib/data/generif-metadata.ttl")
+        generif_checksum = "-1"
+        if generif.exists():
```
jnduli: I think we can move this check to the `hash_generif_graph` and return "-1" if it doesn't exist. This prevents duplicating this logic

scripts/index-genenetwork
BonfaceKilz: See cf90b87bab9e for the resolution.

scripts/index-genenetwork
BonfaceKilz: See: ff96efc26c5f

PR #178, opened 2024-07-12
feat: copy addRif script from genenetwork1

jnduli:
# Description

Copy original script from  https://github.com/genenetwork/genenetwork1/blob/master/web/webqtl/maintainance/addRif.py with python3 compatible changes

## DIff
```
╰─$ diff --minimal --ignore-all-space --text ~/gn/genenetwork1/web/webqtl/maintainance/addRif.py scripts/addRif.py
39d38
< from base import webqtlConfig
43c42
<               con = MySQLdb.Connect(db=webqtlConfig.DB_NAME, host=webqtlConfig.MYSQL_SERVER, user=webqtlConfig.DB_USER, passwd=webqtlConfig.DB_PASSWD)
---
>               con = MySQLdb.Connect(db="gn3", host="localhost", user="gn2", passwd="password")
45c44
<               print "You have successfully connected to mysql.\n"
---
>               print("You have successfully connected to mysql.\n")
47c46
<               print "You entered incorrect password.\n"
---
>               print("You entered incorrect password.\n")
67c66
<               line2 = map(string.strip, string.split(line1, "\t"))
---
>               line2 = line1.strip().split("\t")
86c85
<               line2 = map(string.strip, string.split(line1, "\t"))
---
>               line2 = line1.strip().split("\t")
122d120
<

```

## Follow up Actions

- Change mysql credentials to be passed in from a file
- Use csv to open the files
- ...


scripts/addRif.py
```diff
@@ -0,0 +1,126 @@
...
+import string
+import MySQLdb
+import time
+import os
+import sys
+
+path1 = os.path.abspath(os.path.dirname(__file__))
+path2 = path1 + "/.."
+path3 = path1 + "/../../tmp"
+sys.path.insert(0, path2)
```
BonfaceKilz: Something to consider for a future follow up PR is to `pathlib` for this.  Almost perhaps better naming.

BonfaceKilz (changes requested):
Run pylint and mypy against this file.  I'm sure as is right now it will fail in our CI.

scripts/addRif.py
fredmanglis: @BonfaceKilz @jnduli What are `path1`, `path2`, and `path3`?

fredmanglis: Adding a path to the system paths like this, I find, should be extremely rare, and if possible, avoided. I usually make the script directory a package and use then would do something like `…/python3 -m scripts.addRif [ARGS&OPTS  …]` at the root of the repository - avoids having to mess with paths like this.

With the paths hard-coded like this, if you were to, for some reason, need to move this script to say a sub-package, or elsewhere in the repo, the path system you've set up above will very likely break. On the other hand, the `…/python3 -m <package>[.<package>]*.<module>` will just pick up the paths automatically.

fredmanglis:
*nitpick*: Commit 84718e82675a34e35571e164109ed6588c0eb816 bothers me. Same with the naming of scripts like `index-genenetwork` but I don't care enough to allow that to block progress. I do not know whether there's any advantage other than clean names to naming the scripts in this way. If any of you know, I'm all ears.

PR #179, opened 2024-07-12
Use correct ttl-dir path when generating checksums.

BonfaceKilz:
# Description

Despite changing the TTL file for the generif metadata, CI isn't being triggered.  The `hash_generif_graph` function works by looping through the ttl-dir path and generating the md5 checksum.  This PR fixes this by pointing the hash function to the correct directory, instead of a specific file.

### DIFF
--- a/scripts/index-genenetwork
+++ b/scripts/index-genenetwork
@@ -529,7 +529,7 @@ def is_data_modified(xapian_directory: str,
             ])
         # Return a zero exit status code when the data has changed;
         # otherwise exit with a 1 exit status code.
-        generif = pathlib.Path("/var/lib/data/generif-metadata.ttl")
+        generif = pathlib.Path("/var/lib/data/")
         if (db.get_metadata("generif-checksum").decode() == hash_rdf_graph(generif) and
             db.get_metadata("checksums").decode() == checksums):
             sys.exit(1)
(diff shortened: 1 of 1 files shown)


jnduli (approved):

scripts/index-genenetwork
```diff
@@ -240,7 +240,7 @@ def build_rdf_cache(sparql_uri: str, query: str, remove_common_words: bool = Fal
     return smaller_cache
 
 
-def hash_rdf_graph(ttl_dir: pathlib.Path) -> str:
```
fredmanglis, line 243: I like the (arguably) nicer names. I now know (ideally) what the function should be doing without reading the code. Thanks. I should strive for this too.

scripts/index-genenetwork
```diff
@@ -529,7 +529,7 @@ def is_data_modified(xapian_directory: str,
             ])
         # Return a zero exit status code when the data has changed;
         # otherwise exit with a 1 exit status code.
-        generif = pathlib.Path("/var/lib/data/generif-metadata.ttl")
+        generif = pathlib.Path("/var/lib/data/")
```
fredmanglis, line 532: I'm a little worried about this hard-coded path... maybe we should pass this in as an argument. Consider this for future commits.

PR #180, opened 2024-07-15
Update script that updates Generif_BASIC table

jnduli:
# Description
Refactor and clean the `addRif` script. The changes include:

- Download the files using `requests` python library instead of `wget`
- We read the files as `tsv` using python's csv module.
- We use `INSERT IGNORE` instead of performing a `count` and inserting if the count is 0.
- We pass in the sql_uri from the terminal instead of hardcoding it in the code.

# Tested

Run locally using:
`python scripts/update_rif_table.py --sql-uri "mysql://gn2:password@localhost/gn2"`

Example result:

```

MariaDB [gn2]> select * from GeneRIF_BASIC where VersionId=5 limit 2 \G;
*************************** 1. row ***************************
 SpeciesId: 3
     TaxID: 3702
    GeneId: 814572
    symbol:
 PubMed_ID: 17550895
createtime: 2010-01-21 00:00:00
   comment: AtCCMA interacts with AtCcmB to form a novel mitochondrial ABC transporter involved in cytochrome c maturation in Arabidopsis
 VersionId: 5
*************************** 2. row ***************************
 SpeciesId: 3
     TaxID: 3702
    GeneId: 814592
    symbol:
 PubMed_ID: 20682769
createtime: 2010-12-11 11:15:00
   comment: Results identify a pentatricopeptide repeat protein , BIR6, which is involved in splicing of intron 1 of the mitochondrial nad7 transcript.
 VersionId: 5
2 rows in set (0.048 sec)

ERROR: No query specified

```
and testing on dump from prod:

```
╰─$ python scripts/update_rif_table.py --sql-uri "mysql://gn2:password@localhost/gn2"
Generif_BASIC table updated. Added 230070. Skipped 1274050 because they already exists. In case of error, you can use VersionID=5 to find rows inserted with this script

```
# Note to Reviewers

- I'm assuming `VersionId` is used to track what inserted the data. In my case, I've set this to 4, which means if there's an unknown bug in the script, running `DELETE FROM Generif_BASIC WHERE VersionID=4` will drop the data related to the new script.
- The change for `INSERT IGNORE` should really be `INSERT .... ON DUPLICATE UPDATE` since it means we'll always have the new data. However, I'm unwilling to have this now until we're sure the inserts are ok.

### DIFF
--- a/scripts/update_rif_table.py
+++ b/scripts/update_rif_table.py
@@ -21,9 +21,7 @@
 # This module is used by GeneNetwork project (www.genenetwork.org)
 #
 # Created by GeneNetwork Core Team 2010/08/10
-#
-# Last updated by Lei Yan 2011/02/08
-
+# Updated on Lei Yan 2011/02/08
 # created by Lei Yan 02/08/2011
 
 import os
--- a/scripts/update_rif_table.py
+++ b/scripts/update_rif_table.py
@@ -28,103 +28,110 @@
 Script responsible for updating the GenerRIF_BASIC table
 """
 
+import argparse
+import csv
 import os
+import gzip
+import pathlib
 import sys
+from tempfile import TemporaryDirectory
+from typing import Generator, List
+import requests
+
 import MySQLdb
 
-path1 = os.path.abspath(os.path.dirname(__file__))
-path2 = path1 + "/.."
-path3 = path1 + "/../../tmp"
-sys.path.insert(0, path2)
+from gn3.db_utils import database_connection
+
+TAX_IDS = {"10090": 1, "9606": 4, "10116": 2, "3702": 3}
+
+# os.chdir(path3)
 
 
-def fetchrif():
-    """ TODO: break this down into modules """
-    try:
-        con = MySQLdb.Connect(db="gn3", host="localhost", user="gn2", passwd="password")
-        cursor = con.cursor()
-        print("You have successfully connected to mysql.\n")
-    except:
-        print("You entered incorrect password.\n")
-        sys.exit(0)
+def download_file(url: str, dest: pathlib.Path):
+    """Saves the contents of url in dest"""
+    with requests.get(url, stream=True) as resp:
+        resp.raise_for_status()
+        with open(dest, "wb") as f_:
+            for chunk in resp.iter_content(chunk_size=8192):
+                f_.write(chunk)
 
-    tax_ids = {"10090": 1, "9606": 4, "10116": 2, "3702": 3}
-    tax_id_keys = tax_ids.keys()
 
-    os.chdir(path3)
-    print(f"path3: {path3}")
+def read_tsv_file(fname: pathlib.Path) -> Generator[dict]:
+    with gzip.open(fname, mode="rt") as gz_file:
+        reader = csv.DictReader(gz_file, delimiter="\t", quoting=csv.QUOTE_NOTE)
+        yield from reader
+
+
+def parse_gene_info_from_ncbi(fname: pathlib.Path):
     genedict = {}
+    for row in read_tsv_file(fname):
+        if row["#task_id"] not in TAX_IDS:
+            continue
+        genedict[row["GeneID"]] = genedict[row["Symbol"]]
+    return genedict
 
-    os.system("rm -vf gene_info")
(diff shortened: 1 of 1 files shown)


scripts/update_rif_table.py
```diff
@@ -0,0 +1,133 @@
...
+from tempfile import TemporaryDirectory
+from typing import Dict, Generator
+import requests
+
+from gn3.db_utils import database_connection
+
+TAX_IDS = {"10090": 1, "9606": 4, "10116": 2, "3702": 3}
+
+GENE_INFO_URL = "https://ftp.ncbi.nlm.nih.gov/gene/DATA/gene_info.gz"
+GENERIFS_BASIC_URL = "https://ftp.ncbi.nih.gov/gene/GeneRIF/generifs_basic.gz"
```
BonfaceKilz, line 25: *thought:* avoid creating global variables if they are used once.  The point of global variables is to share some (preferably immutable) state across different functions/modules.  In this case,  I see we only ever use these URLs inside one part (when downloading data), so I recommend you have just directly pass the URL to the function; perhaps with a comment should you feel the need to add one.

scripts/update_rif_table.py
```diff
@@ -0,0 +1,133 @@
...
+                        symbol,  # symbol
+                        row["PubMed ID (PMID) list"],  # PubMed_ID
+                        row["last update timestamp"],  # createtime
+                        row["GeneRIF text"],  # comment
+                        row["#Tax ID"],  # TaxID
+                        VERSION_ID,  # VersionId
+                    )
+                    cursor.execute(insert_query, insert_values)
+        print(
+            f"Generif_BASIC table updated. In case of error, you can do use VersionID={VERSION_ID} to find rows inserted with this script"
```
BonfaceKilz: *nitpick:* fix the grammar in this message.
```suggestion
            f"Generif_BASIC table updated. In case of error, you can use VersionID={VERSION_ID} to find rows inserted with this script"
```

scripts/update_rif_table.py
```diff
@@ -0,0 +1,133 @@
...
+# at rwilliams@uthsc.edu and xzhou15@uthsc.edu
+#
+#
+#
+# This module is used by GeneNetwork project (www.genenetwork.org)
+#
+# Created by GeneNetwork Core Team 2010/08/10
+# Updated on Lei Yan 2011/02/08
+# created by Lei Yan 02/08/2011
+
```
BonfaceKilz: *suggestion (non-blocking):* Consider removing this front-matter.

Some of the information here is old, like the URL link to the sourceforge instance.  We already have the main licensing info in the main repo as well.

*suggestion (non-blocking):* Consider having:

```
#! /usr/bin/env python3
```

at the very top of this script.  That way, when running "./update_rif_table.py" from the terminal, your terminal emulator can automatically pick python3 without the need for you explicitly provide it.

BonfaceKilz (changes requested):

scripts/update_rif_table.py
jnduli: I view these as more like constants, here's the PEP for constants: https://peps.python.org/pep-0008/#constants

It also provides an easy place to change the endpoints in future.

scripts/update_rif_table.py
jnduli: Done

scripts/update_rif_table.py
jnduli: Cool. I've dropped the front-matter and added the shebang.

BonfaceKilz (approved):

BonfaceKilz:
I'll run this by @pjotrp before I run this on TUX02.

robwwilliams:
Idea is to ensure that our GeneRIFs include a minimum of three species:
human, mouse, and rat. Gets a bit tricky due to nomenclature of gene
symbols. Discussion on this topic from Rob via email about 20 days ago.
There are databases at ENSEMBL and NCBI and UCSC that align genes symbols
across many species.

Rob

Robert W Williams PhD
Chair: Department of Genetics, Genomics and Informatics
UT-ORNL Governor's Chair in Computational Genomics
71 S Manassas St, Memphis TN 38163
University of Tennessee Health Science Center, TSRB 405
Cell: 901 604 4752
EMAIL: ***@***.***
Zoom: https://tennesseehipaa.zoom.us/j/9492269658
----------------

PR #186, opened 2024-08-28
Api/fetch the latest wiki by versionid

BonfaceKilz:
# Description

This PR updates the wiki endpoint to fetch the latest entry by versionId.  Similarly, [this](https://git.genenetwork.org/gn-transform-databases/commit/?id=397745b554e03fa2df0784c0c48ac43d01428980) removes blank-nodes in the RDF transform.  Querying blank-nodes adds a layer of complexity we don't need.

### DIFF
--- a/gn3/api/metadata_api/wiki.py
+++ b/gn3/api/metadata_api/wiki.py
@@ -65,7 +65,7 @@ def edit_wiki(comment_id: int):
                                           insert_dict["versionId"], cat_id)
             )
         return jsonify({"success": "ok"})
-    return jsonify(error="Error editting wiki entry, most likely due to DB error!"), 500
+    return jsonify(error="Error editing wiki entry, most likely due to DB error!"), 500
 
 
 @wiki_blueprint.route("/<string:symbol>", methods=["GET"])
--- a/gn3/db/rdf.py
+++ b/gn3/db/rdf.py
@@ -59,7 +59,8 @@ CONSTRUCT {
            dct:created ?created
 } WHERE {
     ?symbolId rdfs:comment _:node ;
-              rdfs:label '$symbol' .
+              rdfs:label ?symbolName .
+    FILTER ( LCASE(?symbolName) = LCASE('$symbol') ) .
     _:node rdf:type gnc:GNWikiEntry ;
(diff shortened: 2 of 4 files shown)


gn3/api/metadata_api/wiki.py
```diff
@@ -64,8 +64,8 @@ def edit_wiki(comment_id: int):
                 category_addition_query, (comment_id,
                                           insert_dict["versionId"], cat_id)
             )
-        return jsonify({"success": "ok"})
-    return jsonify(error="Error editting wiki entry, most likely due to DB error!"), 500
+        return jsonify({"success": "ok"}), 201
```
jnduli: nitpick: I don't mind this but the api we call is `edit_wiki` and I think the expectation is that this modifies  the comment. The fact that we add an extra row is an implementation detail since we could end up changing this to run an `update` instead. Can we do `200` instead?

gn3/db/rdf.py
```diff
@@ -41,40 +41,57 @@ def query_and_frame(query: str, context: dict, endpoint: str) -> dict:
...
-           rdfs:comment ?wikientry .
-    OPTIONAL { _:node gnt:reason ?reason } .
+    ?symbolId rdfs:label ?symbolName .
+    ?uid rdfs:comment ?comment ;
+         gnt:symbol ?symbolId ;
+         rdf:type gnc:GNWikiEntry ;
+         dct:created ?createTime .
+    FILTER ( LCASE(?symbolName) = LCASE('$symbol') ) .
+    {
+        SELECT (MAX(?vers) AS ?max) ?id_ WHERE {
```
jnduli, line 70: I'm not too sure about this approach performance wise. If this behaves similar to sql, we won't be able to optimize this since the subquery will have to run n times. Can we try this approach instead:
```
SELECT distinct_cols, MAX(?vers), other_cols FROM tables GROUP BY distinct_cols
```
I've seen a query that may do something similar here: https://en.wikibooks.org/wiki/SPARQL/Aggregate_functions

BonfaceKilz: Something to consider in another PR.

PR #188, opened 2024-08-30
Chores/merge gnqa api

Alexanderlacuna:
This PR merges gnqa api branch to the main branch.
Fixes merge conflicts.


Alexanderlacuna:
@jnduli @BonfaceKilz  Created the branch to merge to main for gn3 . Kindly review.

tests/unit/test_llm.py
jnduli, line 1: Why have we dropped this test file? I don't see an alternative for the tests.

gn3/api/llm.py
```diff
@@ -1,128 +1,117 @@
...
-        })
-    except Exception as error:
-        return jsonify({"query": query, "error": f"Request failed-{str(error)}"}), 500
-
-
-@GnQNA.route("/rating/<task_id>", methods=["POST"])
+    fahamu_token = current_app.config.get("FAHAMU_AUTH_TOKEN")
+    if not fahamu_token:
+        raise LLMError(
+            "Request failed:an LLM authorisation token  is required ", query)
```
jnduli: minor: `Request failed: an ...` (extra space after :)

gn3/api/llm.py
```diff
@@ -1,128 +1,117 @@
...
+    response = {
+        "task_id": task_id,
+        "query": query,
+        "answer": answer,
+        "references": refs
+    }
+    with (db.connection(current_app.config["LLM_DB_PATH"]) as conn,
+          require_oauth.acquire("profile user") as token):
+        cursor = conn.cursor()
+        cursor.execute("""CREATE TABLE IF NOT EXISTS
```
jnduli, line 40: What's the difference between this query and the query found in `llm_db_tables.sql`?

gn3/api/llm.py
```diff
@@ -1,128 +1,117 @@
...
+        history(user_id TEXT NOT NULL,
+        task_id TEXT NOT NULL,
+        query  TEXT NOT NULL,
+        results  TEXT,
```
jnduli: Can we use `JSONB` for results instead of here? See: https://www.sqlite.org/json1.html

gn3/api/llm.py
```diff
@@ -1,128 +1,117 @@
...
+        created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+        PRIMARY KEY(task_id)) WITHOUT ROWID""")
+        cursor.execute(
+            """INSERT INTO history(user_id, task_id, query, results)
+            VALUES(?, ?, ?, ?)
+            """, (str(token.user.user_id), str(task_id["task_id"]),
+                  query,
+                  json.dumps(response))
+        )
+    return response
```
jnduli, line 54: why not `jsonify(response)`?

gn3/api/llm.py
```diff
@@ -1,128 +1,117 @@
...
-    except Exception as error:
-        return jsonify({"query": query, "error": f"Request failed-{str(error)}"}), 500
-
-
-@GnQNA.route("/rating/<task_id>", methods=["POST"])
+    fahamu_token = current_app.config.get("FAHAMU_AUTH_TOKEN")
+    if not fahamu_token:
+        raise LLMError(
+            "Request failed:an LLM authorisation token  is required ", query)
+    task_id, answer, refs = get_gnqa(
```
jnduli, line 29: why don't we try to pick up some results from the previous query if we already have something in the history? Something like:
```
SELECT answer FROM history WHERE created_at > DATE('now','-1 day')`
```
Of course we'd have to play with the delta time we want.

gn3/api/llm.py
```diff
@@ -1,128 +1,117 @@
...
+    """Api endpoint for rating GNQA query and answer"""
+    with (require_oauth.acquire("profile") as token,
+          db.connection(current_app.config["LLM_DB_PATH"]) as conn):
+        results = request.json
+        user_id, query, answer, weight = (token.user.user_id,
+                                          results.get("query"),
+                                          results.get("answer"),
+                                          results.get("weight", 0))
+        cursor = conn.cursor()
+        create_table = """CREATE TABLE IF NOT EXISTS Rating(
```
jnduli, line 69: is there a difference between this and the tables we define in the sql files?

Also on table design, we also have the History table. Is there a link between these two? I'd assume we'd have the actual queries and answers in the history table, and here we'd only have the ratings.

gn3/api/llm.py
```diff
@@ -1,128 +1,117 @@
...
+              answer TEXT NOT NULL,
+              weight INTEGER NOT NULL DEFAULT 0,
+              task_id TEXT NOT NULL UNIQUE,
+              created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+              PRIMARY KEY(task_id))"""
+        cursor.execute(create_table)
+        cursor.execute("""INSERT INTO Rating(user_id, query,
+        answer, weight, task_id, created_at)
+        VALUES(?, ?, ?, ?, ?, ?)
+        ON CONFLICT(task_id) DO UPDATE SET
```
jnduli, line 81: Given this conflict, I'd rather we have the `user_id` as part of the Primary Key. This was, we can have other users rate the same results (this is especially true when we decide to use caches for similar queries)

gn3/api/llm.py
```diff
@@ -1,128 +1,117 @@
...
+        weight=excluded.weight
+        """, (str(user_id), query, answer, weight, task_id,
+              datetime.now(timezone.utc)))
```
jnduli: let's drop the `datetime.now` and have sqlite automatically add CURRENT_TIMESTAMP. This helps ensure  consistency with how we also handle create_time in the history table.

gn3/api/llm.py
```diff
@@ -1,128 +1,117 @@
...
         return {
-                "message": "You have successfully rated this query:Thank you!!"
-            }, 200
-    except sqlite3.Error as error:
-        return jsonify({"error": str(error)}), 500
-    except Exception as error:
-        raise error
+           "message": "You have successfully rated this query.Thank you!"
```
jnduli: minor: `this query. Thank you!` (space after .)

gn3/api/llm.py
```diff
@@ -1,128 +1,117 @@
...
-        })
-
-
-@GnQNA.route("/get_hist_names", methods=["GET"])
-@handle_errors
-def fetch_prev_hist_ids():
-
-    with (Redis.from_url(current_app.config["REDIS_URI"], decode_responses=True)) as redis_conn:
-        return jsonify({"prev_queries": get_user_queries("random_user", redis_conn)})
+def fetch_prev_history():
```
jnduli: This api does a lot of things that we can break up into different calls, especially since the return types are slightly different. Here's my suggestion:

1. `fetch_previous_user_task_ids`: returns list of task_ids user has searched using
2. `fetch_previous_search(task_id)`: returns list of previous searches using the task_id
3. `delete_search`: deletes single search by a user
4. `delete_searches`: takes in a list of task_ids and deletes them all. (I'm still unsure if this should be a DELETE call though)

gn3/llms/client.py
```diff
@@ -1,72 +1,61 @@
...
-        gnqa = GeneNetworkQAClient(account="account-name", api_key="XXXXXXXXXXXXXXXXXXX...")
+        gnqa = GeneNetworkQAClient(account="account-name",
+    api_key="XXXXXXXXXXXXXXXXXXX...")
     """
 
-    BASE_URL = 'https://genenetwork.fahamuai.com/api/tasks'
-
-    def __init__(self, account, api_key, version="v3", timeout=30, total_retries=5, backoff_factor=30):
+    def __init__(self, api_key, timeout=30,
+                 total_retries=5, backoff_factor=30):
```
jnduli: Question: the defaults here means that our searches will wait for at most 16 minutes before returning something. Is this expected behaviour? I'd expected this to be less. See [backoff_factor explanation](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry) here for how I did my calculations

gn3/llms/client.py
```diff
@@ -80,141 +69,59 @@ def __init__(self, account, api_key, version="v3", timeout=30, total_retries=5,
...
-
-    @staticmethod
-    def ask_the_documents(extend_url, my_auth):
+    def get_answer_using_task_id(self, extend_url, my_auth):
+        """call this method with task id to fetch response"""
         try:
-            response = requests.post(
-                base_url + extend_url, data={}, headers=my_auth)
+            response = requests.get(
+               self.answer_url + extend_url, data={}, headers=my_auth)
```
jnduli: 1. can we use urljoin e.g. `urljoin(self.answer_url, extended_url)` incase we have trailing and leading `/`
2. If we are using `headers=my_auth`, why do we have `self.headers` defined in the __init__ function?

gn3/llms/client.py
```diff
@@ -80,141 +69,59 @@ def __init__(self, account, api_key, version="v3", timeout=30, total_retries=5,
...
+    def ask(self, ex_url, query,  *args, **kwargs):
+        """fahamu ask api interface"""
+        self.query = query
+        res = self.custom_request('POST', f"{self.base_url}{ex_url}", *args, **kwargs)
+        return res, json.loads(res.text)
 
     def get_answer(self, taskid, *args, **kwargs):
-        query = self.answer_url + self.extendTaskID(taskid)
+        """Fahamu get answer interface"""
+        query = f"{self.answer_url}?task_id={taskid['task_id']}"
```
jnduli: `taskid['task_id']`? Can we rename `taskid` to something else?

gn3/llms/client.py
```diff
@@ -80,141 +69,59 @@ def __init__(self, account, api_key, version="v3", timeout=30, total_retries=5,
...
         res = self.custom_request('GET', query, *args, **kwargs)
-        if (res.status_code != 200):
-            return self.negative_status_msg(res), 0
         return res, 1
 
     def custom_request(self, method, url, *args, **kwargs):
-
+        """ make custom request to fahamu api ask and get response"""
```
jnduli: can we add a comment here explaining why we aren't using the Adapter for Retries we defined in the constructor?

gn3/llms/process.py
```diff
@@ -1,40 +1,74 @@
...
-# pylint: disable=C0301
+BASEDIR = os.path.abspath(os.path.dirname(__file__))
+
+
+class DocIDs():
+    """ Class Method to Parse document id and names from files"""
+    def __init__(self):
+        """
+        init method for Docids
+        * doc_ids.json: opens doc)ids for gn references
```
jnduli: typo in `doc)ids` and `doci_ids`?

gn3/llms/process.py
```diff
@@ -43,32 +77,23 @@ def parse_context(context, get_info_func, format_bib_func):
...
 def load_file(filename, dir_path):
-    """function to open and load json file"""
+    """Utility function to read json file
+    Args:
+        filename:  file name to read
+        dir_path:  base directory for the file
+    Returns: json data read to a dict
+    """
     file_path = os.path.join(dir_path, f"{filename}")
     if not os.path.isfile(file_path):
```
jnduli: I don't think we need this check though. By default `open` will raise the error. Here's a test I did:
```

In [7]: open("/tmp/adb.sdaf", "rb")
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
Cell In [7], line 1
----> 1 open("/tmp/adb.sdaf", "rb")

FileNotFoundError: [Errno 2] No such file or directory: '/tmp/adb.sdaf'

```

gn3/llms/process.py
```diff
@@ -1,40 +1,74 @@
...
+class DocIDs():
+    """ Class Method to Parse document id and names from files"""
+    def __init__(self):
+        """
+        init method for Docids
+        * doc_ids.json: opens doc)ids for gn references
+        * sugar_doc_ids:  open doci_ids for diabetes references
+        """
+        self.doc_ids = load_file("doc_ids.json", BASEDIR)
+        self.sugar_doc_ids = load_file("all_files.json", BASEDIR)
```
jnduli: I don't think we use `sugar_doc_ids` outside this class, but I may be wrong. If not, we can change this to:
```
sugar_doc_ids = load_file("all_files.json', BASEDIR)
self.format_doc_ids(sugar_doc_ids)
```

gn3/llms/process.py
```diff
@@ -1,40 +1,74 @@
...
+           doc_id
+        Args:
+            doc_id: str: a search key for doc_ids
+        Returns:
+              an object with doc_info if doc_id in doc_ids
+        """
+        if doc_id in self.doc_ids.keys():
+            return self.doc_ids[doc_id]
+        else:
+            return doc_id
```
jnduli: can we raise an exception here instead? And in the calling function, we catch this and handle this use_case?

gn3/llms/process.py
```diff
@@ -43,32 +77,23 @@ def parse_context(context, get_info_func, format_bib_func):
         doc_info = get_info_func(doc_ids)
```
jnduli: Follow up to https://github.com/genenetwork/genenetwork3/pull/188/files#r1740678411
```
try:
  doc_info = get_info_func(doc_ids)
  bib_info = format_bib_func(doc_info)
except KeyError:
  bib_info = doc_ids
```

gn3/llms/process.py
jnduli, line 1: I think we can add unit tests for some functionality defined here

gn3/api/llm.py
Alexanderlacuna: During implementation, we decided to treat each query from each user as unique, even if the queries are identical. This approach results in storing multiple entries, each identified by a distinct task ID. The goal of this strategy is to allow users to modify their ratings. Users, users will only have access to their own results.

tests/unit/test_llm.py
Alexanderlacuna: I  found the tests to be trivial and therefore unnecessary.

gn3/llms/process.py
Alexanderlacuna: This commit fixes: https://github.com/genenetwork/genenetwork3/pull/188/commits/742beb6ee663bc9ae5409461d2be4b2144b8893e

gn3/llms/process.py
Alexanderlacuna: commit fixes this: https://github.com/genenetwork/genenetwork3/pull/188/commits/086c80510ff418bca77f544d3dd4b174d2dc9c8e

gn3/llms/process.py
Alexanderlacuna: commit fixes this: https://github.com/genenetwork/genenetwork3/pull/188/commits/2e81e48695e9b5618746c8cd1c6c83b452836442

gn3/api/llm.py
Alexanderlacuna: commit fixes this: https://github.com/genenetwork/genenetwork3/pull/188/commits/884f57de47dfb7e80fd8ff25760ce5353267964e

gn3/api/llm.py
Alexanderlacuna: Commit fixes this: https://github.com/genenetwork/genenetwork3/pull/188/commits/748ebc5baff3412df163e17ee18a8fc688329b1d

gn3/api/llm.py
Alexanderlacuna: This is cause we treat each query as independent.

gn3/api/llm.py
Alexanderlacuna: ooh! I see what you mean

Alexanderlacuna: checked this and flask casts this to jsonify(response)   for dict.https://github.com/pallets/flask/blob/d091bb00c0358e9f30006a064f3dbb671b99aeae/src/flask/app.py#L2017

gn3/llms/process.py
Alexanderlacuna: Noted.

gn3/api/llm.py
Alexanderlacuna: There is no difference between the tables created for the LLM SQL file and those created in the files. However, not all records are in the rating table, and it also misses some references

gn3/api/llm.py
Alexanderlacuna: No difference

gn3/llms/client.py
Alexanderlacuna: Added docs on why this is the case:
See,https://github.com/genenetwork/genenetwork3/pull/188/commits/3de74bf43245088ff4d07af4fd796eb510ff73bb

gn3/llms/client.py
Alexanderlacuna: This adapter is not  used anywhere but that needs to fixed to a better default.

gn3/llms/client.py
Alexanderlacuna: fixes by: https://github.com/genenetwork/genenetwork3/pull/188/commits/303fa9ed617ff843d663719dd76d3a08c00cf724

gn3/api/llm.py
Alexanderlacuna: Cool, I suggest to implement these changes in a different PR together with adding test for the process file.

gn3/llms/process.py
Alexanderlacuna: See ref:https://github.com/genenetwork/genenetwork3/pull/188#discussion_r1740958728

gn3/llms/process.py
Alexanderlacuna: Am quite unsure why you prefer this approach  doesn't that make  the get_info_func just a dict  retrieval  with no other functionality which could be defined in the caller function

gn3/llms/client.py
Alexanderlacuna: The entire function is from some older code no longer in use :: Removed.
https://github.com/genenetwork/genenetwork3/pull/188/commits/41c352926ca178e65d6c948fdf93b0f987e2878a

gn3/api/llm.py
Alexanderlacuna: `Noted  to address this in the fixup PR.`

gn3/api/llm.py
Alexanderlacuna: fixed.

Alexanderlacuna:
@jnduli These are the changes to be resolved in the fixup PR.
* Adding tests to the process file.
* Using Jsonb for results
* split the history api endpoint

gn3/api/llm.py
jnduli: Noted. Let's visit this when we handle caching ai results

gn3/llms/process.py
jnduli: Yeah. I'd prefer it this way since we end up being explicit if an entry exists or not. With the current approach, we have to do a `if doc_ids == doc_info` here https://github.com/genenetwork/genenetwork3/pull/188/files#r1740680853; which means we need some means of handling this.

Having an exception means other users for this api don't have to remember to do this check.

gn3/llms/process.py
Alexanderlacuna: Commit fixes this: https://github.com/genenetwork/genenetwork3/pull/188/commits/c16c54759cfd493250424ee3f565862e5d6009b3

jnduli (approved):

PR #191, opened 2024-09-12
Add Caching Support Using History Table for LLM search

jnduli:
## Description

We need to support caching to reduce the dependence on LLM queries. Since we already have a history table, I've updated the search query to use this to fetch repeat results from this table.

Constraints:

- the query has to be done by the same user (not ideal but a good temporary solution)
- similar queries done within the last 24 hour period will return caching. This may limit testing/development when we upload the papers. We can set this to be configurable later though.

Tested:

See screenshots here https://github.com/genenetwork/genenetwork2/pull/867

### DIFF
--- a/gn3/api/llm.py
+++ b/gn3/api/llm.py
@@ -15,35 +15,72 @@ from gn3.auth import db
 
 gnqa = Blueprint("gnqa", __name__)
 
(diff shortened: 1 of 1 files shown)


Alexanderlacuna (approved):
LGTM!

PR #193, opened 2024-09-17
Api/edit rif endpoint

BonfaceKilz:
## Description

This PR extends the current RIF edit end-points to allow editing in RDF.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,10 @@ These configurations should be set in an external config file, pointed to with t
 - SPARQL_ENDPOINT (ex: "http://localhost:9082/sparql")
 - XAPIAN_DB_PATH (ex: "/export/data/genenetwork/xapian")
 - TMPDIR
+- SPARQL_USER
+- SPARQL_ENDPOINT (ex: "http://localhost:9082/sparql-auth/")
+- SPARQL_PASSWORD
+- SPARQL_AUTH_URI
 
 TMPDIR also needs to be set correctly for the R script(s) because they pass results on as files on the local system (previously there was an issue with it being set to /tmp instead of ~/genenetwork3/tmp). Note that the Guix build system should take care of the paths.
 
--- a/gn3/api/metadata_api/wiki.py
+++ b/gn3/api/metadata_api/wiki.py
@@ -25,16 +25,16 @@ def edit_wiki(comment_id: int):
     # FIXME: attempt to check and fix for types here with relevant errors
     payload: Dict[str, Any] = request.json  # type: ignore
     pubmed_ids = [str(x) for x in payload.get("pubmed_ids", [])]
-
+    created = datetime.datetime.now(datetime.timezone.utc).strftime(
+        "%Y-%m-%d %H:%M:%S"
+    )
     insert_dict = {
         "Id": comment_id,
         "symbol": payload["symbol"],
         "PubMed_ID": " ".join(pubmed_ids),
         "comment": payload["comment"],
         "email": payload["email"],
-        "createtime": datetime.datetime.now(datetime.timezone.utc).strftime(
-            "%Y-%m-%d %H:%M"
-        ),
+        "createtime": created,
         "user_ip": request.environ.get("HTTP_X_REAL_IP", request.remote_addr),
         "weburl": payload.get("web_url"),
         "initial": payload.get("initial"),
@@ -78,6 +78,7 @@ def edit_wiki(comment_id: int):
             comment_id=comment_id,
             payload=payload,
             next_version=next_version,
+            created=created,
             sparql_conf={
                 "sparql_user": current_app.config["SPARQL_USER"],
                 "sparql_password": current_app.config["SPARQL_PASSWORD"],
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -124,6 +124,10 @@ These configurations should be set in an external config file, pointed to with t
 - SPARQL_ENDPOINT (ex: "http://localhost:9082/sparql")
 - XAPIAN_DB_PATH (ex: "/export/data/genenetwork/xapian")
 - TMPDIR
+- SPARQL_USER
+- SPARQL_ENDPOINT (ex: "http://localhost:9082/sparql-auth/")
```
jnduli, line 128: - `SPARQL_AUTH_ENDPOINT` because we already have another `SPARQL_ENDPOINT`?
- does sparql endpoint support having credentials in the url e.g. `http://username@password:localhost....` ref https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#access_using_credentials_in_the_url (If this works it reduces the number of config variables)

gn3/db/rdf/wiki.py
```diff
@@ -180,3 +187,112 @@ def get_comment_history(
     # See note above in the doc-string
     results["data"] = sorted(data, key=lambda d: d["version"], reverse=True)
     return results
+
+
+def get_next_comment_version(
```
jnduli: what should be the source of truth for this function? There's a similar function for this [here](https://github.com/jnduli/genenetwork3/blob/52cd294c2f1d06dddbd6ff613b11f8bc43066038/gn3/db/wiki.py#L51) and I'd rather we stick to one. This prevents drifts when problems occur.

gn3/db/rdf/wiki.py
```diff
@@ -180,3 +187,112 @@ def get_comment_history(
...
+rdf:type gnc:GNWikiEntry ;
+gnt:symbol "$symbol" ;
+dct:identifier "$comment_id"^^xsd:integer ;
+dct:hasVersion "$next_version"^^xsd:integer ;
+dct:created "$created"^^xsd:datetime .
+""").substitute(
+        comment=payload["comment"],
+        name=name, symbol=payload['symbol'],
+        comment_id=comment_id, next_version=next_version,
+        created=datetime.datetime.now(datetime.timezone.utc).strftime('%Y-%m-%d %H:%M:%S'))
```
jnduli: this will be different from the value we store in mysql. Can we make it the same value?

gn3/api/metadata_api/wiki.py
```diff
@@ -66,6 +70,18 @@ def edit_wiki(comment_id: int):
                 category_addition_query, (comment_id,
                                           insert_dict["versionId"], cat_id)
             )
+
+        # Editing RDF:
+        update_wiki_comment(
+            comment_id=comment_id,
+            payload=payload,
```
jnduli: Can we pass over `insert_dict` instead to this child process? I don't like passing in the payload since `insert_dict` would be what we want to insert after doing clean up. We can modify it to work both in mysql and in rdf

README.md
BonfaceKilz: That `SPARQL_AUTH_ENDPOINT` because that is what Virtuoso provides for INSERTS/UPDATES/DELETES operations in a graph.  To use that end-point, you have to provide a user name and a password.

gn3/db/rdf/wiki.py
```diff
@@ -180,3 +187,112 @@ def get_comment_history(
...
+} $using
+""").substitute(prefix=RDF_PREFIXES,
+                graph=graph,
+                comment_triple=comment_triple,
+                using=using),
+        sparql_user=sparql_conf["sparql_user"],
+        sparql_password=sparql_conf["sparql_password"],
+        sparql_auth_uri=sparql_conf["sparql_auth_uri"],
+    )
+    return (res, 200)
```
jnduli: why are we returning 200 here?

gn3/db/rdf/__init__.py
```diff
@@ -126,66 +127,88 @@
...
+        "chromosome": "gnt:chr",
+        "mb": "gnt:mb",
+        "peakLocation": "gnt:locus",
+        "species": "gnt:belongsToSpecies",
+        "group": "gnt:belongsToGroup",
+    }
+)
+
+
+def sparql_query(query: str, endpoint: str, format_type="json-ld") -> dict:
```
jnduli, line 166: minor question: can we set the format_type to be types supported by sparqlwrapper? I was thinking a uniont type between `JSON | JSONLD`, see https://sparqlwrapper.readthedocs.io/en/latest/SPARQLWrapper.Wrapper.html

gn3/api/metadata_api/wiki.py
BonfaceKilz: I've had to add extra keys to the `insert_dict`.

gn3/db/rdf/wiki.py
BonfaceKilz: No good reason.  I reckon we can just return the `res` only.

jnduli (approved):

PR #196, opened 2024-09-23
Handle missing GN3_SECRETS for CI testing.

BonfaceKilz:
This PR handles setting `GN3_SECRETS` for CI testing which currently [fails](https://ci.genenetwork.org/jobs/genenetwork3/631).

In CI Testing for GN3, we don't set any environment.  We run tests defined in the package defined [here](https://github.com/genenetwork/genenetwork3/blob/main/.guix/genenetwork3-package.scm).

### DIFF
--- a/tests/fixtures/rdf.py
+++ b/tests/fixtures/rdf.py
@@ -9,9 +9,20 @@ import pytest
 def get_sparql_auth_conf() -> dict:
     """Fetch SPARQL auth configuration for the GN3_SECRETS file."""
     sparql_conf = config.Config("")
-    # When loading from the environment, GN3_CONF precedes
-    # GN3_SECRETS.  Don't change this order.
-    sparql_conf.from_envvar("GN3_CONF")
+    if os.environ.get("GN3_CONF"):
+        # Check whether GN3_CONF has been set, and ignore GN3_CONF
+        # otherwise.  In CD, we use a mixed-text file, so we don't
+        # have an explicit PATH to point this to.
+        # https://git.genenetwork.org/gn-machines/tree/genenetwork-development.scm#n517
+        sparql_conf.from_envvar("GN3_CONF")
+    # Set sane defaults for GN3_SECRETS to CD's secret file.  In CD,
+    # this file is set in the genenetwork3 cd gexp:
+    # https://git.genenetwork.org/gn-machines/tree/genenetwork-development.scm#n516
+    # However, during testing GN3_SECRETS isn't set; and by default,
+    # we run guix's default tests for python projects: `pytest`
+    # https://git.genenetwork.org/guix-bioinformatics/tree/gn/packages/genenetwork.scm#n182
+    if not os.environ.get("GN3_SECRETS"):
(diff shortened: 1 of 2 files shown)


tests/fixtures/rdf.py
```diff
@@ -9,9 +9,17 @@
 def get_sparql_auth_conf() -> dict:
```
jnduli, line 9: 1. Isn't it dangerous to link to CD's secret file? An alternative would be to use a testing virtuoso instance, which ensures our testing environment is clean.
2. Can we drop the GN3_CONF loading? I think we should enforce having the secrets in the `GN3_SECRETS` file
3. Do we plan to set GN3_SECRETS path in the gexp? If so, let's add a FIXME comment here about removing the `if` statement.

BonfaceKilz: > 1. Isn't it dangerous to link to CD's secret file?

I don't understand what you mean by dangerous?  The location of that file is already exposed in gn-machines.  Unless you have SSH root Access, you can't touch that file.

> An alternative would be to use a testing virtuoso instance, which ensures our testing environment is clean.

This requires bootstrapping that testing environment in guix g-exp.  That looks like alot of work to get things running, yet it's just easier and simpler to use the existing virtuoso instance; and use a testing named-graph which we currently do.

> 2. Can we drop the GN3_CONF loading?

Let's leave this here...

> I think we should enforce having the secrets in the GN3_SECRETS file

... Right now in our gn-machines code, we have:

```
(setenv "GN3_CONF"
        #$(mixed-text-file "gn3.conf"
                           "SPARQL_ENDPOINT=\"" sparql-endpoint "\"\n"
                           "DATA_DIR=\"" data-directory "\"\n"
                           "XAPIAN_DB_PATH=\"" xapian-db-path "\"\n"
                           "AUTH_DB=\"" auth-db-path "\"\n"
                           "LLM_DB_PATH=\"" llm-db-path "\"\n"))
```

I don't know how guix sets path for a mixed-text-file, or how it's handled (probaly in the store).  For this PR, I abuse GN3_SECRETS to set "SPARQL_ENDPOINT".  Either that, or I hard-code a sane-default in the fixture module.   And by default, "SPARQL_ENDPOINT" goes to GN3_CONF.

> 3. Do we plan to set GN3_SECRETS path in the gexp?

No.

> If so, let's add a FIXME comment here about removing the if statement.

GN3_CONF is already set in CD's genenetwork3-cd-gexp:

```
(setenv "GN3_SECRETS" #$gn3-secrets)
```

But in the actual tests themselves, we use the test-phase of:

```
(define-public genenetwork3
  (package
    (inherit gn:genenetwork3)
    (version (string-append (package-version gn:genenetwork3) "-git"))
    (source (local-file ".."
                        "genenetwork3-checkout"
                        #:recursive? #t
                        #:select? (or (git-predicate (dirname (current-source-directory)))
                                      (const #t))))))
```

... to run the tests.  Instead of re-writing the test-phase of genenetwork3 or alternatively creating a g-exp specifically to set GN3_CONF (and override the test-pages), just configuring where to look for it here, and allow developers to configure it locally works fine without being a time-sink to do alot of the extra bootstrapping required.

PR #197, opened 2024-09-24
Rif edit atomicity

jnduli:
## Description

Make rif edit queries atomic by ensuring that mysql transactions are rolled back if RDF update fails. The implementation catches `Exception` because we want any failure from the RDF update to cause the rollback.

Other changes include:

- Bug fix because the return type from InnoDB isn't a string but bytes. This fixes the error `TypeError: Object of type bytes is not JSON serializable`
- RDF query fix, resulting from the error: `Virtuoso 22023 Error SL001: The SPARQL 1.1 LCASE() function needs a string value as 1st argument`

### DIFF
--- a/gn3/api/metadata_api/wiki.py
+++ b/gn3/api/metadata_api/wiki.py
@@ -74,13 +74,17 @@ def edit_wiki(comment_id: int):
                                           insert_dict["versionId"], cat_id)
             )
 
-        # Editing RDF:
-        update_wiki_comment(
-            insert_dict=insert_dict,
-            sparql_user=current_app.config["SPARQL_USER"],
-            sparql_password=current_app.config["SPARQL_PASSWORD"],
-            sparql_auth_uri=current_app.config["SPARQL_AUTH_URI"]
-        )
+        try:
+            # Editing RDF:
+            update_wiki_comment(
+                insert_dict=insert_dict,
+                sparql_user=current_app.config["SPARQL_USER"],
+                sparql_password=current_app.config["SPARQL_PASSWORD"],
+                sparql_auth_uri=current_app.config["SPARQL_AUTH_URI"]
+            )
+        except Exception as exc:
+            conn.rollback()
+            raise exc
         return jsonify({"success": "ok"})
     return jsonify(error="Error editing wiki entry, most likely due to DB error!"), 500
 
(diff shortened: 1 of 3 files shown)


gn3/db/wiki.py
```diff
@@ -9,6 +9,16 @@ class MissingDBDataException(Exception):
     """Error due to DB missing some data"""
 
 
+def _decode_dict(result: dict):
+    new_result = {}
+    for k, v in result.items():
```
BonfaceKilz: We have a pylint error here:

```
************* Module gn3.db.wiki
gn3/db/wiki.py:14:11: C0103: Variable name "v" doesn't conform to snake_case naming style (invalid-name)
```

I would suggest:

```suggestion
    for k, val in result.items():
```

gn3/api/metadata_api/wiki.py
```diff
@@ -74,13 +74,17 @@ def edit_wiki(comment_id: int):
...
+        try:
+            # Editing RDF:
+            update_wiki_comment(
+                insert_dict=insert_dict,
+                sparql_user=current_app.config["SPARQL_USER"],
+                sparql_password=current_app.config["SPARQL_PASSWORD"],
+                sparql_auth_uri=current_app.config["SPARQL_AUTH_URI"]
+            )
+        except Exception as exc:
+            conn.rollback()
```
BonfaceKilz: *issue:*  This raises the following mypy error:

```
gn3/api/metadata_api/wiki.py:86: errogn3/api/metadata_api/wiki.py:86: error: "Connection" has no attribute "rollback"  [attr-defined]r: "Connection" has no attribute "rollback"  [attr-defined]
```
```suggestion
            conn.rollback()  # type: ignore
```

BonfaceKilz (changes requested):
@jnduli the only minor fixes are pylint and mypy errors.  I've tested the minor RDF Query change and all the tests pass.  Make requested changes and this is good to go!

gn3/db/wiki.py
jnduli: Done

gn3/api/metadata_api/wiki.py
jnduli: Done.

Alexanderlacuna:
LGTM!

Alexanderlacuna (approved):

PR #198, opened 2024-09-30
Run tests against Virtuoso that is spun locally.

BonfaceKilz:
# Description

This PR adds virtuoso to the background, and runs tests against it.  The genenetwork3-package is also modified to run rdf tests.

## Testing

Make sure `#:use-module ((gnu packages linux) #:select (procps))` and `#:use-module ((gnu packages databases) #:select (virtmuoso-ose))` are installed in your guix system.

`pytest -k rdf`

Testing the genenetwork3-package:

- Replace `genenetwork3` with `genenetwork3-all-tests` at the bottom of `.guix/genenetwork3-package.scm` file.
- Run:
`guix build -L ~/projects/guix-past -L ~/projects/guix-bioinformatics -f genenetwork3-package.scm`

### DIFF
--- a/.guix/genenetwork3-package.scm
+++ b/.guix/genenetwork3-package.scm
@@ -3,6 +3,8 @@
                 #:select (genenetwork3) #:prefix gn:)
   #:use-module ((gnu packages check) #:select (python-pylint))
   #:use-module ((gnu packages python-check) #:select (python-mypy))
+  #:use-module ((gnu packages linux) #:select (procps))
+  #:use-module ((gnu packages databases) #:select (virtuoso-ose))
   #:use-module (guix gexp)
   #:use-module (guix utils)
   #:use-module (guix git-download)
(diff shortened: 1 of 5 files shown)


tests/fixtures/rdf.py
```diff
@@ -1,30 +1,55 @@
...
+from flask import config
 
 import pytest
 import requests
 from requests.auth import HTTPDigestAuth
 
 
-def get_sparql_auth_conf(config_obj) -> dict:
-    """Fetch SPARQL auth configuration from the configuration object."""
+def get_sparql_auth_conf() -> dict:
```
jnduli: Can we investigate a better way of passing these in? These if statements may cause problems in the future. Some suggestions:

1. Change the return type from a dict to a python obj e.g. namedtuple or dataclass. This ensures stricter handling of the sparql auth items.
2. Instead of having 4 different params that we save in a configuration file, can we instead create one of the from `user@password:uri_def. We follow a similar pattern in our other configs e.g. https://github.com/jnduli/genenetwork3/blob/52cd294c2f1d06dddbd6ff613b11f8bc43066038/gn3/db_utils.py#L13
3. Since we're loading data into virtuoso from start, can we also launch a custom virtuoso instance here using `subprocess`?

BonfaceKilz:
@jnduli I've worked on launching a custom virtuoso instance in our build instance.  We run tests against that.  The extra complexity to bootstrap things is worth it IMHO.

tests/fixtures/rdf.py
```diff
@@ -1,46 +1,96 @@
...
+        return [int(pid) for pid in response.split()]
+
+
+@pytest.fixture(scope="session")
+def rdf_setup():
     """Upload RDF to a Virtuoso named graph"""
-    # Define the URL and file
-    sparql_conf = get_sparql_auth_conf(fxtr_app_config)
-    url = sparql_conf["sparql_crud_auth_uri"]
+    dir_path = os.path.dirname(__file__).split("fixtures")[0]
```
jnduli: minor: `Pathlib.path(__file__).parent.parent`

.guix/genenetwork3-package.scm
```diff
@@ -3,38 +3,51 @@
...
+	         (("virtuoso-t")
+	          (string-append #$virtuoso-ose "/bin/virtuoso-t"))))))
+      	   (add-after 'build 'rdf-tests
+	   	   (lambda _
+	   	     (invoke "pytest" "-k" "rdf")))
+     	   (add-before 'build 'pylint
+      	     (lambda _
+      	       (invoke "pylint" "main.py" "setup.py" "wsgi.py" "setup_commands"
+      	     	      "tests" "gn3" "scripts" "sheepdog")))
+      	   (add-after 'pylint 'mypy
```
Alexanderlacuna, line 44: Question:  Could you explain how this works in the build phase with `add-after`, and why we not invoke  `add-before` for everything?

BonfaceKilz: @Alexanderlacuna Great question!  `add-before` and `add-after` essentially serve the same purpose, but the key difference is where in the build process you want to insert a new step.  In this case, we use `add-after` because we want to run `mypy` right `after pylint` but right before running the unit tests.  It’s a logical flow: we perform our static checks (like pylint and mypy) before running the unit tests to catch issues earlier.  As for the RDF tests, I’ll adjust them to run after the unit tests to maintain that order.

BonfaceKilz: I've addressed above [here](https://github.com/genenetwork/genenetwork3/pull/199)

PR #200, opened 2024-10-08
Api/ncbi metadata

BonfaceKilz:
This PR:
- Creates new test NCBI entries to `tests/test_data/ttl-files/test-data.ttl`
- Add test cases for the RDF queries that fetch NCBI test data
- Creates a new endpoint for fetching NCBI metadata: `GET metadata/rif/<symbol>`

## Testing

Running the new tests:

`pytest -k rdf`

Testing the end-points:

`curl localhost:8080/api/metadata/rif/Lpl`

`curl localhost:8080/api/metadata/rif/Lpl --header "Accept: application/ld+json"`


gn3/db/rdf/wiki.py
```diff
@@ -236,11 +241,83 @@ def update_wiki_comment(
...
+            "taxonomic_id": "skos:notation",
+            "gene_id": "gnt:hasGeneId",
+            "pubmed_id": "dct:references",
+            "created": "dct:created",
+            "comment": "rdfs:comment",
+            "version": "dct:hasVersion",
+        }
+    }
+    data: list[dict[str, int | str]] = []
+    for entry in sparql_query(query=query, endpoint=sparql_uri, format_type="json"):
```
jnduli, line 313: does setting the `format_type` to `json` mean we don't support `application/ld+json` here? The api call seems to support this [here](https://github.com/genenetwork/genenetwork3/pull/200/files#diff-a171a2593fc05b5a1c72c33bd5ebbdd49b80d7c0cd93d8033cdfd9951fe70e84R171) so I wonder how we'll handle this.

BonfaceKilz: This is a `sparqlwrapper` implementation detail.  Using sparqlwrapper and a CONSTRUCT, you get a `json-ld`.  For SELECTs you don't get that.  Instead, here I want a `json`.

PR #204, opened 2024-12-13
Fix/rqtl permutation

Alexanderlacuna:
PR fixes bug for rqtl by fetching the correct value from the output file which has  the following columns:
snp, chromosome, position, lod_score


BonfaceKilz (approved):
LGTM!

PR #225, opened 2025-05-29
Add genotype matrix script.

BonfaceKilz:


scripts/lmdb_matrix.py
```diff
@@ -0,0 +1,164 @@
...
+def create_database(db_path: str) -> lmdb.Environment:
+    """Create or open an LMDB environment."""
+    return lmdb.open(db_path, map_size=100 * 1024 * 1024, create=True)
+
+
+def genotype_db_put(db: lmdb.Environment, genotype: GenotypeMatrix) -> bool:
+    metadata = json.dumps(genotype.metadata).encode("utf-8")
+    with db.begin(write=True) as txn:
+        txn.put(b"matrix", genotype.matrix.tobytes())
+        txn.put(b"transpose", genotype.transpose.tobytes())
```
Alexanderlacuna: Question: Why do you need to store the  transposed matrix?

scripts/lmdb_matrix.py
```diff
@@ -0,0 +1,164 @@
...
+                key, value = line[1:].split(':', 1)
+                file_metadata[key] = value
+            else:
+                stream.seek(stream.tell() - len(line) - 1)
+                break
+
+        # Read header
+        header = stream.readline().strip().split()
+        metadata_columns = ["Chr", "Locus", "cM",
+                            "Mb"] if "Mb" in header else ["Chr", "Locus", "cM"]
```
Alexanderlacuna: suggestion:  why not use a unified structure for all files with all the provided headers and using a value such as 'N/A' when a header value is missing?

scripts/lmdb_matrix.py
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        for key in metadata_columns[2:]:
+            metadata[key] = []
+
+        locus, chromosomes = [], []
+        for i in range(nrows):
+            line = stream.readline().strip().split()
+            meta, data = line[:len(metadata_columns)
+                              ], line[len(metadata_columns):]
```
Alexanderlacuna, line 306: Suggestion: Could you add a comment for this just to make it clear what you are fetching?
i.e # this returns chr ... etc

scripts/lmdb_matrix.py
```diff
@@ -0,0 +1,164 @@
...
+            for j, element in enumerate(data):
+                if element.isdigit():
```
Alexanderlacuna: Question:  
Since the objective of this is to obtain the chromosome value, what happens with the sex chromosomes which is a string, for example, the 'X' chromosome ?

scripts/lmdb_matrix.py
```diff
@@ -0,0 +1,164 @@
...
+                    matrix[i, j] = int(element)
+                elif element == maternal:
+                    matrix[i, j] = 0
+                elif element == paternal:
```
Alexanderlacuna: suggestion: I think this if else block to get  geno_codes is too rigid,
how about something like:
```py
  genotype_map = {
    maternal: 0,
    paternal: 1,
    heterozygous: 2,
    unknown: 3
}
value = value = genotype_map.get(element)
matrix[i, j ] =value
```
With this you can extend the number of genotype categories, you can also  add that as a part of the metadata

Alexanderlacuna:
@BonfaceKilz  is this task complete?

scripts/lmdb_matrix.py
BonfaceKilz: I've removed this in my updated work.  You are right, we don't need this for now; since that adds un-necessary overhead for a really huge matrix.

scripts/lmdb_matrix.py
BonfaceKilz: But they are that rigid lolz.  Regardless, `match ... case` FTW.

Alexanderlacuna (approved):
LGTM!

PR #231, opened 2025-07-11
feat: Implement gnqna rate limiter.

Alexanderlacuna:
This PR  implements rate limiter for gnqna.

### DIFF
--- a/gn3/api/llm.py
+++ b/gn3/api/llm.py
@@ -1,8 +1,11 @@
 """Api endpoints for gnqa"""
+import ipaddress
 import json
 import string
 import uuid
+
 from datetime import datetime
+from datetime import timedelta
 from typing import Optional
 from functools import wraps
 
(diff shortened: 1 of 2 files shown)


gn3/api/llm.py
```diff
@@ -104,6 +118,74 @@ def call_with_anonymous_fallback():
...
+
+def check_rate_limiter(request_metadata, db_path, tokens_lifespan=1440, default_tokens=4):
+    """
+    Checks if an anonymous user has a valid token within the given lifespan.
+    If expired or not found, creates or resets the token bucket.
+    `tokens_lifespan` is in seconds.  24*60
+    default_token set to 4 requests per hour.
+    """
+    # Extract IP address /identifier
+    user_metadata = json.loads(request_metadata.headers.get("Anony-Metadata", {}))
```
BonfaceKilz: *issue:* here pass in the dict, instead of processing the request object.  This way, when testing for this specific function will be easier.  Also, I want you to add test cases for the following scenarios: 

- First-time visitor
- Visitor at limit
- Expired token bucket
- Malformed or missing IP metadata

BonfaceKilz (changes requested):
Consider making the rate_limiting function a decorator.  It's nice when we have:

```
@gnqa.route("/search", methods=["GET"])
@with_gnqna_fallback
@require_token
@rate_limit(token_lifespan=1440)
def search(auth_token=None, valid_anony=False):
    [...]
```

Regardless, before I merge this, I want to add tests for the rate limit function before I merge.  See comment above.

gn3/api/llm.py
Alexanderlacuna: @BonfaceKilz I’ve made the changes you requested, except for converting it into a decorator.
I’m okay with users hitting the cache   (since it’s not expensive) — the rate limiter is only intended to control requests to the Fahamu API resource. Turning this into a decorator would also require moving the cache logic into it, which could complicate the flow. Too many decorators make the API logic harder to follow.

